PR #7, opened 2022-10-17
Feature/cleanup

nklomp:
- Cleanups
- Make sure that actual steps can be performed using the output from the previous step
- README updates
- Bugfixes
- Fix packaging

Mainly the way we are creating proofs will need to be refactored at one point. We are basically exposing payloads as options there and we should have more logic aboud DIDs/JWKs etc in there. But that is for a later version


lib/AccessTokenClient.ts
```diff
@@ -0,0 +1,139 @@
...
+  ErrorResponse,
+  GrantTypes,
+  IssuanceInitiationRequestPayload,
+  IssuanceInitiationWithBaseUrl,
+  IssuerTokenEndpointOpts,
+} from './types';
+
+export class AccessTokenClient {
+  // private _clientId?: string;
+  // private _authorizationServerUrl?: string;
```
maikel-maas, line 18: Remove commented out code

lib/AccessTokenClient.ts
```diff
@@ -0,0 +1,139 @@
...
+      isPinRequired: issuanceInitiationRequest.user_pin_required || false,
+      issuerOpts: { issuer: issuanceInitiationRequest.issuer },
+      asOpts: opts?.asOpts ? { ...opts.asOpts } : undefined,
+    };
+    return await this.acquireAccessTokenUsingRequest(await this.createAccessTokenRequest(issuanceInitiationRequest, clientId, opts), reqOpts);
+  }
+
+  public async acquireAccessTokenUsingRequest(
+    accessTokenRequest: AccessTokenRequest,
+    opts: { isPinRequired?: boolean; asOpts?: AuthorizationServerOpts; issuerOpts?: IssuerTokenEndpointOpts }
```
maikel-maas, line 36: Create a new type of it

lib/CredentialRequestClient.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  public async acquireCredentialsUsingProof(
+    proof: ProofOfPossession | ProofOfPossessionOpts,
+    opts?: {
+      credentialType?: string | string[];
+      format?: CredentialFormat | CredentialFormat[];
+      overrideIssuerURL?: string;
+      overrideAccessToken?: string;
+    }
```
maikel-maas, line 40: Create a new type of it

lib/CredentialRequestClient.ts
```diff
@@ -0,0 +1,84 @@
...
+  ): Promise<CredentialResponse | ErrorResponse> {
+    const request = await this.createCredentialRequest(proof, { ...opts });
+    return await this.acquireCredentialsUsingRequest(request, { ...opts });
+  }
+
+  public async acquireCredentialsUsingRequest(
+    request: CredentialRequest,
+    opts?: { overrideIssuerURL?: string; overrideAccessToken?: string }
```
maikel-maas, line 48: Make a type of it

lib/CredentialRequestClient.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+    return { ...responseJson } as CredentialResponse;
+  }
+
+  public async createCredentialRequest(
+    proof: ProofOfPossession | ProofOfPossessionOpts,
+    opts?: {
+      credentialType?: string | string[];
+      format?: CredentialFormat | CredentialFormat[];
+    }
```
maikel-maas, line 68: Make a type of it

lib/types/Authorization.types.ts
```diff
@@ -0,0 +1,82 @@
...
+export interface IssuerTokenEndpointOpts {
+  issuer: string;
+  tokenEndpoint?: string;
+}
+
+export interface AccessTokenRequestOpts {
+  asOpts?: AuthorizationServerOpts;
+  pin?: number; // Pin-number. Only used when required
+  // client_id?: string;
+}
```
maikel-maas, line 33: Remove commented-out code

maikel-maas (changes requested):
Left some small remarks

lib/IssuanceInitiation.ts
```diff
@@ -0,0 +1,28 @@
+import { convertJsonToURI, convertURIToJsonObject } from './functions';
+import { IssuanceInitiationRequestPayload, IssuanceInitiationWithBaseUrl } from './types';
+
+export class IssuanceInitiation {
```
hrehman-sphereon, line 4: It seems more like a Util perhaps it should be renamed and similar to HTTP Utils. And perhaps it can go in that folder. 

Also functions seems incorrect when it contains the classes ...

tests/CredentialRequestClient.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+  publicKey: undefined,
+};
+
+beforeAll(async () => {
+  const keyPair = await jose.generateKeyPair('ES256');
+  jwtArgs.privateKey = keyPair.privateKey as KeyObject;
+  jwtArgs.publicKey = keyPair.publicKey as KeyObject;
+});
+
+describe('VcIssuanceClient ', () => {
```
hrehman-sphereon, line 54: 'CredentialRequestClient '

renaming is required.

tests/CredentialRequestClient.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+  it('should build correctly provided with correct params', function () {
+    const vcIssuanceClient = CredentialRequestClient.builder()
```
hrehman-sphereon, line 56: Perhaps a comprehensice replace All needs to run .

vcIssuanceClient needs to be renamed.

tests/CredentialRequestClient.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+      .build();
+    const proof: ProofOfPossession = await createProofOfPossession({
+      issuerURL: vcIssuanceClient.getIssuerUrl(),
+      jwtSignerArgs: jwtArgs,
+      jwtSignerCallback: (args) => signJWT(args),
+    });
+    const credentialRequest: CredentialRequest = await vcIssuanceClient.createCredentialRequest(proof);
+    expect(credentialRequest.proof.jwt.includes(partialJWT)).toBeTruthy();
+    const result: ErrorResponse | CredentialResponse = await vcIssuanceClient.acquireCredentialsUsingRequest(credentialRequest);
+    expect(result['credential']).toBeDefined();
```
hrehman-sphereon, line 126: if possible can we have equality check instead of it being defined.

tests/IssuanceInitiation.spec.ts
```diff
@@ -0,0 +1,141 @@
+import { AuthzFlowType, convertJsonToURI, convertURIToJsonObject } from '../lib';
+import { IssuanceInitiation } from '../lib';
```
hrehman-sphereon, line 2: It was suggested to rename the source file. If the source file is renamed. then this must also be renamed.

tests/IssuanceInitiation.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+  it('should return pre-authorized code flow with a credential_type array', () => {
+    expect(
+      AuthzFlowType.valueOf({
+        issuer: 'test',
+        credential_type: ['test', 'test1'],
+        pre_authorized_code: 'test',
+      })
+    ).toEqual(AuthzFlowType.PRE_AUTHORIZED_CODE_FLOW);
+  });
+});
```
hrehman-sphereon, line 141: The relevant code seemed to be moved to different source class. And these test cases should be moved to a different file.

hrehman-sphereon (changes requested):
Some small comments.

hrehman-sphereon:
@nklomp Where have we coppied the ProofUtil From ? Is that written for this particular project ?
If there is already some project from which we copied this code. Perhaps test cases can also be copied.

lib/AccessTokenClient.ts
nklomp: Could you explain to my why?

lib/IssuanceInitiation.ts
nklomp: Nah I don't want to call everything a meaningless Util. That literally is saying nothing

This is a core class for interaction of the developer. As you can see it is using functions with some params, so that we do not have to ask the developer to know with which arguments the function should be called to begin with. Why would that go into a Utils folder?

tests/IssuanceInitiation.spec.ts
nklomp: Nah. For an external Developer having a single class about IssuanceInitiation makes way more sense

lib/AccessTokenClient.ts
nklomp: To be clear I can think of one reason. Just want to hear why you believe this needs to be done

maikel-maas: It makes it reusable and it will fit the screen, making the code look cleaner. If it was Java I would say that it would improve readability, in Typescript I'm not really sure.

tests/IssuanceInitiation.spec.ts
nklomp: Proof Util was part of the Credential Request Client, but that class was doing too many things. As creating the proof should also be possible using your own code, I moved these to functions.

The whole project needs better test coverage to begin with, but that is not what this refactoring and fixing was about. First we need to get code which is relatively easy to use and comprehend by external developers.

lib/AccessTokenClient.ts
maikel-maas: Please share your reason, so I can learn something from it.

lib/IssuanceInitiation.ts
hrehman-sphereon: There is a directory named functions. If this class were to move there ... we would need to rename that directory. But since we are not moving I think that name is fine.

lib/AccessTokenClient.ts
nklomp: The only reason is consistency to be honest.
Readability really depends on the actual values. Why are we okay for instance with a function like this:

````typescript
randomFunction(arg1: string, arg2:number)
````

Why don't we say this needs an Interface for the arguments? Simply because that hides the arguments. Upon first glance you cannot see anymore what the actual arguments will be. It could be a very complex object, whilst in reality it is only 2 arguments.

In the original line for you comment it is debatable whether adding an interface for the opts is increasing readability. Right now you can see that there is a boolean , together with optional Authorization Server Options and Issuer Token Endpoints Options. At least you get a feel for what the arguments will be about. If we replace that with a singular Interface, you certainly cannot deduce what the options are about, without actually looking at the interface.

I roughly take as a rule for myself that if a function or argument object gets more than 3 params, to move it to an interface to improve readability.

There are always exceptions possible of course. For instance whenever you have cases where arguments or options can be re-used across different methods. Then it makes sense to immediately move them to an interface for re-usability.

lib/AccessTokenClient.ts
nklomp: removed

lib/types/Authorization.types.ts
nklomp: removed

tests/CredentialRequestClient.spec.ts
nklomp: Renamed

tests/IssuanceInitiation.spec.ts
nklomp: Moved

nklomp:
Comments should be addressed

PR #8, opened 2022-10-19
feature/JFF-15/server_metadata

maikel-maas:
Had problems with the Formats type, had to change it to a mapped type to limit the input for the keys and even so I was getting an error during the build.

### DIFF
--- a/lib/types/ServerMetadata.ts
+++ b/lib/types/ServerMetadata.ts
@@ -38,15 +38,20 @@ export interface Claims {
   };
 }
 
-export interface Formats {
-  [x: FormatType]: {
+export enum FormatKeys {
+  ldp_vc = 'ldp_vc',
+  jwt_vc = 'jwt_vc',
+  mdl_iso = 'mdl_iso',
+  ac_vc = 'ac_vc',
+}
+
+export type Formats = {
+  [x in FormatKeys]: {
     types: string[];
     cryptographic_binding_methods_supported?: string[];
     cryptographic_suites_supported?: string[];
     [x: string]: unknown;
   };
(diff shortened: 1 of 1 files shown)


lib/types/ServerMetadata.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export interface Logo {
+  url?: string;
+  alt_text?: string;
+  description?: string;
+  background_color?: string;
+  text_color?: string;
+  [x: string]: unknown;
+}
```
hrehman-sphereon, line 29: As per specs this should be in Display interface. And example is different ... I guess let's approve it.

hrehman-sphereon (approved):

maikel-maas:
@hrehman-sphereon This is part of logo:

logo: OPTIONAL. A JSON object with information about the logo of the Credential with a following non-exhaustive list of parameters that MAY be included:

url: OPTIONAL. URL where the Wallet can obtain a logo of the Credential issuer.

alt_text: OPTIONAL. String value of an alternative text of a logo image.

description: OPTIONAL. String value of a description of the Credential.

background_color: OPTIONAL. String value of a background color of the Credential represented as numerical color values defined in CSS Color Module Level 37 [[CSS-Color](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#CSS-Color)].

text_color: OPTIONAL. String value of a text color of the Credential represented as numerical color values defined in CSS Color Module Level 37 [[CSS-Color](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#CSS-Color)].

PR #9, opened 2022-10-20
Hotfix/token form post

nklomp:
We should not use a normal post but use a form post. Also allowed form some additional default and custom headers


hrehman-sphereon (approved):
LGTM ...

PR #10, opened 2022-10-21
Feature/jff 15 server metadata client

nklomp:
Additiona of metadata client to allow fetching metadata from the issuer and AS


maikel-maas (approved):
It looks good to me

maikel-maas (approved):
It looks good to me

PR #12, opened 2022-10-28
added proof builder

sksadjad:
added proof builder
refactored tests


lib/functions/HttpUtils.ts
```diff
@@ -19,15 +19,15 @@ export async function getJson<T>(URL: string): Promise<T> {
   } else {
     if (response.status && response.status < 400) {
       const json = await response.json();
-      console.log(`Well-knonw response: ${JSON.stringify(json, null, 2)}`);
+      // console.log(`Well-knonw response: ${JSON.stringify(json, null, 2)}`);
```
maikel-maas, line 22: Remove commented out code

nklomp: No please don't.  We will do that at the end. We really need to be able to see responses and/or enable comments at this stage

lib/functions/HttpUtils.ts
```diff
@@ -19,15 +19,15 @@ export async function getJson<T>(URL: string): Promise<T> {
...
       return json as T;
```
maikel-maas, line 23: Casting to T?

lib/functions/ProofUtils.ts
```diff
@@ -0,0 +1,80 @@
...
+  };
+}
+
+function partiallyValidateJWS(jws: string): void {
+  if (jws.split('.').length !== 3 || !jws.startsWith('ey')) {
+    throw new Error(JWS_NOT_VALID);
+  }
+}
+
+function setJWSDefaults(args: ProofOfPossessionCallbackArgs, issuerUrl: string, clientId?: string): { header: unknown; payload: unknown } {
```
maikel-maas, line 40: What about using `payload: JWTPayload`?

lib/functions/ProofUtils.ts
```diff
@@ -0,0 +1,80 @@
...
+    iat: args.payload['iat'] ? args.payload['iat'] : now / 1000 - 60, // Let's ensure we subtract 60 seconds for potential time offsets
+    exp: args.payload['exp'] ? args.payload['exp'] : now / 1000 + 10 * 60,
+  };
+  const defaultHeader: JWTHeaderParameters = {
+    alg: 'ES256',
+    typ: 'JWT',
+  };
+  args.payload = { ...defaultPayload, ...(args.payload as unknown as ProofOfPossessionCallbackArgs) };
+  args.header = { ...defaultHeader, ...(args.header as unknown as any) };
+  return args as unknown as any;
```
maikel-maas, line 62: I think you can use `JWT types` here or return `any` directly if needed, with `// eslint-disable-next-line @typescript-eslint/no-explicit-any` on top or it.

lib/functions/ProofUtils.ts
```diff
@@ -0,0 +1,80 @@
...
+      aud: popSignInputDecodedArgs.aud,
+      iss: popSignInputDecodedArgs.iss,
+      iat: popSignInputDecodedArgs.iat,
+      exp: popSignInputDecodedArgs.exp,
+      jti: popSignInputDecodedArgs.jti,
+    },
+    '',
+    {
+      alg: popSignInputDecodedArgs.signAlgorithm as unknown as any,
+      typ: popSignInputDecodedArgs.type as unknown as any,
```
maikel-maas, line 77: I think you can use JWT types here, return any directly if needed, with // eslint-disable-next-line @typescript-eslint/no-explicit-any on top or it or cast it to string

lib/types/CredentialIssuance.types.ts
```diff
@@ -48,57 +46,60 @@ export type EncodeJsonAsURIOpts = { uriTypeProperties?: string[]; arrayTypePrope
 
 export type DecodeURIAsJsonOpts = { requiredProperties?: string[]; arrayTypeProperties?: string[] };
 
-export interface JWK {
```
maikel-maas, line 51: this is a subset present in the specs. I think it's better than a generic unknown

lib/types/CredentialIssuance.types.ts
```diff
@@ -48,57 +46,60 @@ export type EncodeJsonAsURIOpts = { uriTypeProperties?: string[]; arrayTypePrope
...
+  iat: number;
+  iss: string;
+  jti?: string;
+  jwk?: string;
+  x5c?: string;
+  kid?: string;
+  nonce: string;
+  signAlgorithm?: 'EdDSA' | string;
+  type?: 'JWT' | string;
+  [x: string]: unknown;
```
maikel-maas, line 73: I don't think this is correct. these properties are related to JWT and not proof of purpose, besides you are mixing the header and payload.

tests/CredentialRequestClient.spec.ts
```diff
@@ -1,60 +1,26 @@
...
-  privateKey: undefined,
-  publicKey: undefined,
-};
-
-beforeAll(async () => {
-  const keyPair = await jose.generateKeyPair('ES256');
-  jwtArgs.privateKey = keyPair.privateKey as KeyObject;
-  jwtArgs.publicKey = keyPair.publicKey as KeyObject;
-});
+async function proofOfPossessionCallbackFunction(_args: ProofOfPossessionCallbackArgs): Promise<string> {
```
maikel-maas, line 21: // eslint-disable-next-line @typescript-eslint/no-unused-vars

tests/CredentialRequestClient.spec.ts
```diff
@@ -71,14 +37,20 @@ describe('Credential Request Client ', () => {
...
+          payload: {
+            aud: 'aud',
+            iss: 'sphereon',
+          },
+        },
+      })
+      .build();
     const credentialRequest: CredentialRequest = await credReqClient.createCredentialRequest(proof);
-    expect(credentialRequest.proof.jwt).toContain(partialJWT);
+    expect(credentialRequest.proof.jwt).toContain('ey.val.eu');
```
maikel-maas, line 59: I'm wondering why the actual call was changed to a mocked one. This is a working example to the user.

tests/CredentialRequestClient.spec.ts
```diff
@@ -121,45 +99,60 @@ describe('Credential Request Client ', () => {
...
     const result: ErrorResponse | CredentialResponse = await credReqClient.acquireCredentialsUsingRequest(credentialRequest);
     expect(result['credential']).toEqual(mockedVC);
   });
   it('should fail creating a proof of possession with simple verification', async () => {
-    const credReqClient = CredentialRequestClient.builder()
-      .withCredentialEndpoint('https://oidc4vci.demo.spruceid.com/credential')
-      .withFormat('jwt_vc')
-      .withCredentialType('https://imsglobal.github.io/openbadges-specification/ob_v3p0.html#OpenBadgeCredential')
-      .build();
+    async function proofOfPossessionCallbackFunction(_args: ProofOfPossessionCallbackArgs): Promise<string> {
```
maikel-maas, line 136: // eslint-disable-next-line @typescript-eslint/no-unused-vars

tests/CredentialRequestClient.spec.ts
```diff
@@ -121,45 +99,60 @@ describe('Credential Request Client ', () => {
...
     ).rejects.toThrow(Error(JWS_NOT_VALID));
   });
+
   it('should fail creating a proof of possession with verify callback function', async () => {
-    const credReqClient = CredentialRequestClient.builder()
-      .withCredentialEndpoint('https://oidc4vci.demo.spruceid.com/credential')
-      .withFormat('jwt_vc')
-      .withCredentialType('https://imsglobal.github.io/openbadges-specification/ob_v3p0.html#OpenBadgeCredential')
-      .build();
+    async function proofOfPossessionCallbackFunction(_args: ProofOfPossessionCallbackArgs): Promise<string> {
```
maikel-maas, line 161: // eslint-disable-next-line @typescript-eslint/no-unused-vars

maikel-maas (changes requested):
Left some remarks

lib/functions/HttpUtils.ts
nklomp: Yeah, that is a generic, also not part of this PR. @sksadjad Please leave the logs in place. We remove them all at once. We really need those at this point

lib/functions/HttpUtils.ts
sksadjad: reverted that

package.json
```diff
@@ -26,8 +26,11 @@
   },
   "dependencies": {
     "@sphereon/ssi-types": "0.8.1-next.31",
+    "@types/jwt-encode": "^1.0.0",
```
nklomp, line 29: Why are we adding @types as dep and not as devDep? It is only needed for typing

package.json
```diff
@@ -26,8 +26,11 @@
...
     "bs58": "^5.0.0",
     "cross-fetch": "^3.1.5",
+    "jwt-decode": "^3.1.2",
```
nklomp, line 32: Hmzz, is there a need we now depend on 2 additional JWT libs, whilst we are working with a callback to create the eventual JWTs anyway?

lib/types/CredentialIssuance.types.ts
```diff
@@ -48,57 +46,60 @@ export type EncodeJsonAsURIOpts = { uriTypeProperties?: string[]; arrayTypePrope
...
 }
 
-export interface JWTSignerArgs {
-  header: JWTHeader;
-  payload: JWTPayload;
-  privateKey: KeyObject;
-  publicKey: KeyObject;
+export interface PoPPayloadDecoded {
```
nklomp, line 75: Hmzz, I am not really happy with removing generic well understood concepts like JWT headers, payloads with names that are actually doing something similar, but then tossed together. The original implementation had these nicely seperated

nklomp (changes requested):
See remarks.

nklomp (changes requested):
Both changes are worse than the original tbh.

The original has a clear comment next to it, why it substracted 60 seconds. Clocks are never exactly in sync and since we need to provide a proof of key possession, we get in trouble if the iat is being created one a machine with a clock in the future compared to the verifier. From their perspective the JWT values isn't valid yet.

maikel-maas (approved):
It looks good to me

lib/functions/ProofUtil.ts
```diff
@@ -48,10 +54,10 @@ function setJWSDefaults(args: JWTSignerArgs, issuerUrl: string, clientId?: strin
   const defaultPayload: Partial<JWTPayload> = {
     aud,
     iss,
-    iat: args.payload.iat ? args.payload.iat : now / 1000,
+    iat: args.payload.iat ? args.payload.iat : now / 1000, // Let's ensure we subtract 60 seconds for potential time offsets
```
nklomp, line 57: Where is the actual substraction?

package.json
sksadjad: removed this and used jose

package.json
sksadjad: removed this and used jose

lib/types/CredentialIssuance.types.ts
sksadjad: renamed to the previous version

lib/functions/ProofUtil.ts
sksadjad: changed these two to be:
```
iat: args.payload.iat ? args.payload.iat : now / 1000 - 60, // Let's ensure we subtract 60 seconds for potential time offsets
exp: args.payload.exp ? args.payload.exp : now / 1000 + 10 * 60,
```

lib/ProofOfPossessionBuilder.ts
```diff
@@ -0,0 +1,26 @@
+import { createProofOfPossession } from './functions';
+import { PROOF_CANT_BE_CONSTRUCTED, ProofOfPossession, ProofOfPossessionOpts } from './types';
+
+export class ProofOfPossessionBuilder {
+  popCallbackOpts?: ProofOfPossessionOpts;
+  popEncoded?: ProofOfPossession;
+
+  withPoPCallbackOpts(popCallbackOpts: ProofOfPossessionOpts): ProofOfPossessionBuilder {
```
nklomp, line 8:
```suggestion
  withProofCallbackOpts(proofCallbackOpts: ProofOfPossessionOpts): ProofOfPossessionBuilder {
```

This is already the Proof of Possession builder, so it is logical that the proof is about possession. Sugest to rename it to the above (of course you need to change the param names in the code as well

lib/ProofOfPossessionBuilder.ts
```diff
@@ -0,0 +1,26 @@
...
+    this.popCallbackOpts = popCallbackOpts;
+    return this;
+  }
+
+  withPoPEncoded(popEncoded: ProofOfPossession): ProofOfPossessionBuilder {
```
nklomp, line 13: What does encoded mean in this case? This is simply withProof right?

lib/types/CredentialIssuance.types.ts
```diff
@@ -57,7 +55,30 @@ export interface JWK {
...
+  payload: JWTPayload;
+  [x: string]: unknown;
+}
+
+export interface ProofOfPossessionOpts {
+  clientId?: string;
+  issuerURL?: string;
+  proofOfPossessionCallback: JWTSignerCallback;
+  proofOfPossessionVerifierCallback?: JWTVerifyCallback;
+  proofOfPossessionCallbackArgs: ProofOfPossessionCallbackArgs;
```
nklomp, line 71: Why aren't these anr argument of the actual signer callback?

lib/types/CredentialIssuance.types.ts
```diff
@@ -57,7 +55,30 @@ export interface JWK {
   n?: string;
 }
 
-export type Alg = 'ES256' | 'EdDSA';
+export interface ProofOfPossessionCallbackArgs {
+  privateKey: unknown;
```
nklomp, line 59: Public and private key should not be exposed at all. It is up to the implementation how to handle that. I would expect a kid/did in there though

lib/types/CredentialIssuance.types.ts
```diff
@@ -76,29 +97,7 @@ export interface JWTPayload {
...
-  clientId?: string;
-  jwtSignerArgs: JWTSignerArgs;
-  jwtSignerCallback: JWTSignerCallback;
-  jwtVerifyCallback?: JWTVerifyCallback;
-}
-
-export type JWTSignerCallback = (args: JWTSignerArgs) => Promise<string>;
-
-export type JWTVerifyCallback = (args: JWTVerifyArgs) => Promise<void>;
+export type JWTSignerCallback = (args: { header?: unknown; payload?: unknown; [x: string]: unknown }) => Promise<string>;
```
nklomp, line 100: Why are these unknowns? I would expect something that resembles a JWT header and payload here

lib/ProofOfPossessionBuilder.ts
```diff
@@ -0,0 +1,26 @@
...
+    this.popEncoded = popEncoded;
+    return this;
+  }
+
```
nklomp, line 17: I am missing the functions to set the kid, iss, clientid in the builder from the IRR, Accesstoken response etc?

Please have a look at the integration code in the wallet. Where the Proof is being created I would expect a builder that can set most options from the existing/available objects, together with some additional info like a kid, client_id value. Then it is up to the actual callback to fetch the private key/did and sign the JWT.

nklomp (changes requested):
I am missing the functions to set the kid, iss, clientid in the builder from the IRR, Accesstoken response etc?

Please have a look at the integration code in the wallet. Where the Proof is being created I would expect a builder that can set most options from the existing/available objects, together with some additional info like a kid, client_id value. Then it is up to the actual callback to fetch the private key/did and sign the JWT.

lib/types/CredentialIssuance.types.ts
sksadjad: I've removed this because we have `[x: string]: unknown;` listed.

lib/types/CredentialIssuance.types.ts
sksadjad: added `clientId` and `issuerURL` to `ProofOfPossessionCallbackArgs` properties

lib/ProofOfPossessionBuilder.ts
sksadjad: the values `iss`, `kid`, `jti`, `iat`, and `exp` are handled in `setJWSDefaults`

README.md
```diff
@@ -172,32 +183,29 @@ export interface JWTHeader {
...
 
 export interface ProofOfPossessionOpts {
-  issuerURL?: string;
-  clientId?: string;
-  jwtSignerArgs: JWTSignerArgs;
-  jwtSignerCallback: JWTSignerCallback;
-  jwtVerifyCallback?: JWTVerifyCallback;
+  proofOfPossessionCallback: JWTSignerCallback;
+  proofOfPossessionVerifierCallback?: JWTVerifyCallback;
+  proofOfPossessionCallbackArgs: ProofOfPossessionCallbackArgs;
```
nklomp, line 203: Why do we list the callback args as an option? I would basically expect the build/sign method to have all info present then populate the header/payload and hand it over to the registered callback to get the signed result

lib/functions/ProofUtil.ts
```diff
@@ -7,26 +20,21 @@ import { BAD_PARAMS, JWS_NOT_VALID, JWTHeader, JWTPayload, JWTSignerArgs, ProofO
  *         - jwtSignerCallback: function to sign the proof
```
nklomp, line 20: Doc is not in line with implementation anymore

lib/functions/ProofUtil.ts
```diff
@@ -35,25 +43,34 @@ function partiallyValidateJWS(jws: string): void {
...
   }
-  const iss = args.payload.iss ? args.payload.iss : clientId;
+  if (!args.kid) {
+    throw new Error('No kid provided');
+  }
+  const iss = endpointMetadata.token_endpoint ? endpointMetadata.token_endpoint : args.payload.iss;
   if (!iss) {
     throw new Error('No clientId provided');
   }
+  const jti = args.payload.jti ? args.payload.jti : uuidv4();
```
nklomp, line 62: jti generation should be made optional. So if no jti is provided we should not automaitcally create one. JTI is fully optional

lib/functions/ProofUtil.ts
```diff
@@ -35,25 +43,34 @@ function partiallyValidateJWS(jws: string): void {
...
+  }
+  const aud = endpointMetadata.issuer ? endpointMetadata.issuer : args.payload.aud;
   if (!aud) {
     throw new Error('No issuer url provided');
   }
-  const iss = args.payload.iss ? args.payload.iss : clientId;
+  if (!args.kid) {
+    throw new Error('No kid provided');
+  }
+  const iss = endpointMetadata.token_endpoint ? endpointMetadata.token_endpoint : args.payload.iss;
```
nklomp, line 58: This doesn't make sense. this is the client id that needs to be passed in. It has nothing to do with the token endpoint. 

The client is the issuer/iss of the PoP JWT. That has nothing todo with the credential issuer. The client is creating a signed message that proves they are in control of the keys associated with the kid value

lib/functions/ProofUtil.ts
```diff
@@ -35,25 +43,34 @@ function partiallyValidateJWS(jws: string): void {
...
   const defaultPayload: Partial<JWTPayload> = {
+    jti,
     aud,
     iss,
     iat: args.payload.iat ? args.payload.iat : now / 1000 - 60, // Let's ensure we subtract 60 seconds for potential time offsets
     exp: args.payload.exp ? args.payload.exp : now / 1000 + 10 * 60,
   };
-  const defaultHeader: JWTHeader = {
+  const defaultHeader: JWTHeaderParameters = {
     alg: 'ES256',
```
nklomp, line 71: We need to be able to update the alg value

lib/types/CredentialIssuance.types.ts
```diff
@@ -57,7 +55,26 @@ export interface JWK {
...
+}
+
+export interface ProofOfPossessionOpts {
+  proofOfPossessionCallback: JWTSignerCallback;
+  proofOfPossessionVerifierCallback?: JWTVerifyCallback;
+  proofOfPossessionCallbackArgs: ProofOfPossessionCallbackArgs;
+}
+
+export enum Alg {
+  ES256 = 'ES256',
```
nklomp, line 71: ES256K is missing. Also using a type for what endsup in a payload is a bit more natural. But as long as the enum names are exactly the same as the string value it is okay

README.md
```diff
@@ -172,32 +183,29 @@ export interface JWTHeader {
...
 }
 
-export type JWTSignerCallback = (args: JWTSignerArgs) => Promise<string>;
+export type JWTSignerCallback = (args: { header?: JWTHeader; payload?: JWTPayload; [x: string]: unknown }) => Promise<string>;
```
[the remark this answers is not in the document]
nklomp: Okay guess the documentation needs to be updated from actual current implementation

README.md
nklomp: Okay guess the documentation needs to be updated from actual current implementation

nklomp (changes requested):
see remarks

lib/types/CredentialIssuance.types.ts
sksadjad: added `ES256K`

lib/functions/ProofUtil.ts
sksadjad: changed this to first read from the `args.header.alg` value and if not exist, populate it with `ES256`

lib/functions/ProofUtil.ts
sksadjad: changed this to be completely optional

lib/functions/ProofUtil.ts
sksadjad: Added another field to builder for `clientId`

README.md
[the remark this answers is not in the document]
sksadjad: changed the doc to be inline with our current impl

README.md
sksadjad: changed the name of the interface from `ProofOfPossessionOpts` to `ProofOfPossessionArgs`

lib/functions/ProofUtil.ts
sksadjad: changed the doc

lib/functions/ProofUtil.ts
```diff
@@ -35,13 +53,23 @@ function partiallyValidateJWS(jws: string): void {
...
+  args: ProofOfPossessionCallbackArgs,
+  endpointMetadata: EndpointMetadata,
+  clientId?: string
+): { header: JWTHeader; payload: JWTPayload } {
   const now = +new Date();
-  const aud = args.payload.aud ? args.payload.aud : issuerUrl;
+  if (!endpointMetadata) {
+    throw new Error('No endpointMetadata provided');
+  }
+  const aud = endpointMetadata.issuer ? endpointMetadata.issuer : args.payload.aud;
```
nklomp, line 65: I would expect this the other way around. endpointMetadata always contains an issuer value. 
That should be the default. But if you do supply an issuer as argument use that

lib/functions/ProofUtil.ts
```diff
@@ -35,13 +53,23 @@ function partiallyValidateJWS(jws: string): void {
   }
 }
 
-function setJWSDefaults(args: JWTSignerArgs, issuerUrl: string, clientId?: string): JWTSignerArgs {
+function setJWSDefaults(
```
nklomp, line 56: Suggest to rename this to something like createJWT. Because it is certainly not settings defaults on a JSON Web Signature

lib/functions/ProofUtil.ts
```diff
@@ -35,13 +53,23 @@ function partiallyValidateJWS(jws: string): void {
...
+  }
+  const aud = endpointMetadata.issuer ? endpointMetadata.issuer : args.payload.aud;
   if (!aud) {
     throw new Error('No issuer url provided');
   }
-  const iss = args.payload.iss ? args.payload.iss : clientId;
+  if (!args.kid) {
+    throw new Error('No kid provided');
+  }
+  const iss = clientId ? clientId : args.payload.iss;
```
nklomp, line 72: The iss value should be fully optional and only populated if we actually do use a clientId

lib/ProofOfPossessionBuilder.ts
```diff
@@ -0,0 +1,38 @@
+import { createProofOfPossession } from './functions';
+import { EndpointMetadata, PROOF_CANT_BE_CONSTRUCTED, ProofOfPossession, ProofOfPossessionArgs } from './types';
+
+export class ProofOfPossessionBuilder {
+  clientId?: string;
+  endpointMetadata: EndpointMetadata;
+  proof?: ProofOfPossession;
+  proofCallbackOpts?: ProofOfPossessionArgs;
+
+  withProofCallbackOpts(proofCallbackOpts: ProofOfPossessionArgs): ProofOfPossessionBuilder {
```
nklomp, line 10: Suggest to make this and the withProof static and rename them to fromX. So that it becomes clear you will either start from a proof or from the metatada

lib/types/CredentialIssuance.types.ts
nklomp: Point is that I would not expect the CallbackArgs in the interface to begin with. The callback function you have to implement gets the CallbackArgs passed in. So the implementation needs to be able to accept these arguments. Why would we list these in the ProofOfPossesionArgs? The values are mostly created and supplied to the callback

nklomp (changes requested):
See remarks

lib/functions/ProofUtil.ts
sksadjad: populated after the creation of `jwtArgs` and _if present_

lib/functions/ProofUtil.ts
sksadjad: renamed to `createJWT`

lib/functions/ProofUtil.ts
sksadjad: changed the order in ternary

nklomp (approved):
LGTM

PR #20, opened 2023-01-10
fix: wrong headers type

### LINKED ISSUE #18: Wrong `content-type` is set for `acquireAccessToken`
I came across an issue when trying to acquire the access token (by running `OpenID4VCIClient.acquireAccessToken()`).

I'm currently testing my setup against [MATTR's launchpad](https://launchpad.mattrlabs.com/). When I try to run `OpenID4VCIClient.acquireAccessToken()`, I'm getting a `400` response code with the following error message: `only application/x-www-form-urlencoded content-type bodies are supported on POST /token`.

After monitoring my outbound traffic I found that `content-type` was set to `text/plain;charset=UTF-8`. When I intercept the outgoing request and change the `content-type` to `application/x-www-form-urlencoded` the request succeeds and everything works as expected.

After some debugging, I can't seem to find the problem in this codebase. [HttpUtils.formPost](https://github.com/Sphereon-Opensource/OpenID4VCI-client/blob/d7491baf298ff53f38eb586ab931decb67b3780d/lib/functions/HttpUtils.ts#L15) seems to set the `content-type` correctly... I'm not sure, but it could be an issue with `cross-fetch`.

I will continue looking into this but wanted to post it here in case anyone else runs into the same problem.

Love the API and solid typing by the way, nice job! 🚀

karimStekelenburg:
Changed the fallback `headers` to an empty object instead of array.

Fixes #18 

### DIFF
--- a/lib/functions/HttpUtils.ts
+++ b/lib/functions/HttpUtils.ts
@@ -40,7 +40,7 @@ const openIdFetch = async <T>(
     exceptionOnHttpErrorStatus?: boolean;
   }
 ): Promise<OpenIDResponse<T>> => {
-  const headers = opts?.customHeaders ? opts.customHeaders : [];
+  const headers = opts?.customHeaders ? opts.customHeaders : {};
   if (opts?.bearerToken) {
     headers['Authorization'] = `Bearer ${opts.bearerToken}`;
   }


nklomp:
Ouch, nasty one. Thanks for the PR, will merge and do a release later tonight.

nklomp (approved):
Thanks, nasty one not picked up by the IDE

PR #24, opened 2023-01-11
fix: content-type

karimStekelenburg:
Uncommented this line because it broke the flow and resulted in the same error as mentioned in #18. Not sure if this breaks the Windows fix you did @nklomp...


nklomp:
Hmzz, it was always meant to be re-added again, but with a bit more complexity for handling corner cases. Totally forgot. 
Going to close this PR, as the adjusted code takes the custom headers and content-type argument into account and doesn't set the fallback application/json content-type header at all on a GET request. New version will be released shortly

PR #25, opened 2023-01-11
fix: ssi-types dependency

karimStekelenburg:
The move of the `@sphereon/ssi-types` dependency to `devDependencies` broke my installation as the build includes imports to that package. 


nklomp (approved):
Hmzz, will have a look later where it is used.. Let's revert indeed.

PR #27, opened 2023-01-25
fix: add scope to AccessTokenResponse

karimStekelenburg:
I've noticed that the scope was not present on the `AccessTokenResponse`
type. However, the scope is set whenever using an issuer URI that
contains a `credential_type`.


nklomp (approved):
LGTM

PR #28, opened 2023-03-10
feat: authorization request

karimStekelenburg:
This PR adds functionality to make a request an access token using the [`scope` parameter](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-09.html#name-using-scope-parameter-to-re). Access token requests using the [`authorization_details` parameter](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-09.html#name-request-issuance-of-a-certa) are currently not supported. It also adds a utility method to the `OpenID4VCIClient` class to construct and encode an authorization URL (the URL that points can be used to redirect the user to a login screen).

- [x] Tested
- [ ] Documented

Funded by the Ontario Government


lib/AccessTokenClient.ts
```diff
@@ -80,18 +94,24 @@ export class AccessTokenClient {
...
 
     if (issuanceInitiationRequest[PRE_AUTH_CODE_LITERAL]) {
+      if (codeVerifier) {
+        throw new Error('Cannot pass a code_verifier when flow type is pre-authorized');
+      }
       request.grant_type = GrantTypes.PRE_AUTHORIZED_CODE;
       request[PRE_AUTH_CODE_LITERAL] = issuanceInitiationRequest[PRE_AUTH_CODE_LITERAL];
     }
     if (issuanceInitiationRequest.op_state) {
       this.throwNotSupportedFlow();
```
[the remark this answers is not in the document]
nklomp: Okay guess that part still isn't supported. I think the next comment about moving the check for pre-auth code and auth code is even more relevant to not confuse another developer, seeing code after something which clearly throws an error

lib/types/Authorization.types.ts
```diff
@@ -33,14 +38,27 @@ export interface AccessTokenRequestOpts {
...
 }
 
 export interface AuthorizationRequest {
   response_type: ResponseType.AUTH_CODE;
   client_id: string;
+  code_challenge: string;
+  code_challenge_method: CodeChallengeMethod;
   redirect_uri: string;
-  scope?: string;
+  scope?: string[];
```
nklomp, line 53: A scope is a string which is whitespace separated in case of multiple values
https://www.rfc-editor.org/rfc/rfc6749.html#appendix-A.4

Also from the same spec:
````
 The value of the scope parameter is expressed as a list of space-
   delimited, case-sensitive strings.  The strings are defined by the
   authorization server.  If the value contains multiple space-delimited
   strings, their order does not matter, and each string adds an
   additional access range to the requested scope.
````

Happy for us to make that change to string

File: lib/types/Authorization.types.ts [hunk 3]
```diff
@@ -33,14 +38,27 @@ export interface AccessTokenRequestOpts {
...
+}
+
+export interface AuthorizationRequestOpts {
+  clientId: string;
+  codeChallenge: string;
+  codeChallengeMethod: CodeChallengeMethod;
+  redirectUri: string;
+  scope?: string[];
```
nklomp, line 61: A scope is a string which is whitespace separated in case of multiple values
https://www.rfc-editor.org/rfc/rfc6749.html#appendix-A.4

Since these are arguments, I would expect this to at least accept a string, with maybe an optional string[], the end result however should be that it gets converted to a whitespace separated string

Also from the same spec:
````
 The value of the scope parameter is expressed as a list of space-
   delimited, case-sensitive strings.  The strings are defined by the
   authorization server.  If the value contains multiple space-delimited
   strings, their order does not matter, and each string adds an
   additional access range to the requested scope.
````

Happy for us to make these changes btw

File: lib/types/OpenID4VCIServerMetadata.ts [hunk 4]
```diff
@@ -9,6 +9,7 @@ export interface OpenID4VCIServerMetadata {
   credential_issuer?: CredentialIssuer; //  A JSON object containing display properties for the Credential issuer.
   token_endpoint?: string; //NON-SPEC compliant, but used by several issuers. URL of the OP's Token Endpoint. This URL MUST use the https scheme and MAY contain port, path and query parameter components.
   authorization_server?: string; //NON-SPEC compliant, but used by some issuers. URL of the AS. This URL MUST use the https scheme and MAY contain port, path and query parameter components.
+  authorization_endpoint?: string;
```
nklomp, line 12:
```suggestion
  // TODO: The above authorization_server being used in the wild, serves roughly the same purpose as the below spec compliant endpoint. Look at how to use authorization_server as authorization_endpoint in case it is present
  authorization_endpoint?: string;
```

lib/OpenID4VCIClient.ts
```diff
@@ -70,17 +71,80 @@ export class OpenID4VCIClient {
...
+    if (!scope) {
+      throw Error('Please provide a scope. authorization_details based requests are not supported at this time');
+    }
+
+    if (!this._serverMetadata.openid4vci_metadata.authorization_endpoint) {
+      throw Error('Server metadata does not contain authorization endpoint');
+    }
+
+    // make sure the first scope is 'openid'
+    if (scope.includes('openid')) {
```
nklomp, line 84: See other remarks about this needing to be a space separated string. The order doesn't matter (see other comment)

nklomp (changes requested):
Hi @karimStekelenburg 
Thanks, these are great additions for the authz flow with scopes. 

I have some comments, happy for us to address some if you like

lib/AccessTokenClient.ts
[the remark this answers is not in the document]
karimStekelenburg: Yes I think I've made a mistake because it's unclear what the `op_state` variable is used for. I assumed the presence of `issuanceInitiationRequest[PRE_AUTH_CODE_LITERAL]` would determine the flow type. Am I wrong here?

lib/OpenID4VCIClient.ts
```diff
@@ -70,17 +71,80 @@ export class OpenID4VCIClient {
...
+        scope.unshift('openid');
+      }
+    } else {
+      // if the 'openid' scope isn't present at all, add it
+      scope.unshift('openid');
+    }
+    if (scope.length < 2) {
+      throw Error("Scope array only contains the 'openid' scope. Please also provide a credential type");
+    }
+    console.log(scope);
```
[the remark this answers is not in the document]
karimStekelenburg: Whoops, forgot to remove that by mistake :)

lib/AccessTokenClient.ts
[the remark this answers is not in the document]
nklomp: Yeah, but I guess we will want to refactor that in the future. As the wallet probably will not know upfront which flow it is going to deal with, so we will have to deduct it from the payloads

lib/OpenID4VCIClient.ts
karimStekelenburg: I was converting the array into a space separated string because I thought it was a more convinient API. That being said, staying close to the spec makes total sense. I've refactored it so it now takes a string and injects the 'openid' scope if it's not present. I did leave the parameter itself as optional and throw an error if it's not present though. The reason being that it's optional according to the spec if you use the `authorization_details` parameter instead, but we throw because it's currently not supported.

lib/types/Authorization.types.ts
karimStekelenburg: Done! See other comments for a description.

nklomp (approved):

nklomp:
Thanks for the contribution

PR #32, opened 2023-03-22
feature/VDX-177/authorization_details

maikel-maas:
- AuthorizationDetails type created;
   - Added param to createAuthorizationREquestUrl method;

### DIFF
--- a/packages/client/lib/AuthorizationDetailsBuilder.ts
+++ b/packages/client/lib/AuthorizationDetailsBuilder.ts
@@ -13,17 +13,13 @@ export class AuthorizationDetailsBuilder {
     return this;
   }
 
-  withFormats(format: CredentialFormat | CredentialFormat[]): AuthorizationDetailsBuilder {
-    if (this.authorizationDetails.format && Array.isArray(this.authorizationDetails.format)) {
-      this.authorizationDetails.format.push(...format);
-    } else {
-      this.authorizationDetails.format = format;
-    }
+  withFormats(format: CredentialFormat): AuthorizationDetailsBuilder {
+    this.authorizationDetails.format = format;
     return this;
   }
 
-  withLocations(locations: string | string[]): AuthorizationDetailsBuilder {
-    if (this.authorizationDetails.locations && Array.isArray(this.authorizationDetails.locations)) {
+  withLocations(locations: string[]): AuthorizationDetailsBuilder {
+    if (this.authorizationDetails.locations) {
       this.authorizationDetails.locations.push(...locations);
     } else {
       this.authorizationDetails.locations = locations;
--- a/packages/client/lib/AuthorizationDetailsBuilder.ts
+++ b/packages/client/lib/AuthorizationDetailsBuilder.ts
@@ -29,11 +29,11 @@ export class AuthorizationDetailsBuilder {
 
   addLocation(location: string): AuthorizationDetailsBuilder {
     if (this.authorizationDetails.locations) {
-      this.authorizationDetails.locations.push(location)
+      this.authorizationDetails.locations.push(location);
     } else {
-      this.authorizationDetails.locations = [location]
+      this.authorizationDetails.locations = [location];
     }
-    return this
+    return this;
   }
 
   build(): AuthorizationDetails {
--- a/packages/client/lib/AuthorizationDetailsBuilder.ts
+++ b/packages/client/lib/AuthorizationDetailsBuilder.ts
@@ -27,6 +27,15 @@ export class AuthorizationDetailsBuilder {
     return this;
   }
 
+  addLocation(location: string): AuthorizationDetailsBuilder {
+    if (this.authorizationDetails.locations) {
+      this.authorizationDetails.locations.push(location)
+    } else {
+      this.authorizationDetails.locations = [location]
+    }
+    return this
+  }
+
   build(): AuthorizationDetails {
     if (this.authorizationDetails.format && this.authorizationDetails.type) {
       return this.authorizationDetails as AuthorizationDetails;
--- a/packages/client/lib/AuthorizationDetailsBuilder.ts
+++ b/packages/client/lib/AuthorizationDetailsBuilder.ts
@@ -1,0 +1,40 @@
+import { AuthorizationDetails } from '@sphereon/openid4vci-common';
+import { CredentialFormat } from '@sphereon/ssi-types';
+
+export class AuthorizationDetailsBuilder {
+  private authorizationDetails: Partial<AuthorizationDetails>;
+
+  constructor() {
+    this.authorizationDetails = {};
+  }
+
+  withType(type: string): AuthorizationDetailsBuilder {
+    this.authorizationDetails.type = type;
+    return this;
+  }
+
+  withFormats(format: CredentialFormat | CredentialFormat[]): AuthorizationDetailsBuilder {
+    if (this.authorizationDetails.format && Array.isArray(this.authorizationDetails.format)) {
+      this.authorizationDetails.format.push(...format);
+    } else {
+      this.authorizationDetails.format = format;
+    }
+    return this;
+  }
+
+  withLocations(locations: string | string[]): AuthorizationDetailsBuilder {
+    if (this.authorizationDetails.locations && Array.isArray(this.authorizationDetails.locations)) {
+      this.authorizationDetails.locations.push(...locations);
+    } else {
+      this.authorizationDetails.locations = locations;
+    }
+    return this;
+  }
+
+  build(): AuthorizationDetails {
+    if (this.authorizationDetails.format && this.authorizationDetails.type) {
+      return this.authorizationDetails as AuthorizationDetails;
+    }
+    throw new Error('Type and format are required properties');
+  }
+}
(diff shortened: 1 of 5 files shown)


hrehman-sphereon (approved):
LGTM

packages/client/lib/OpenID4VCIClient.ts
```diff
@@ -71,37 +71,55 @@ export class OpenID4VCIClient {
...
       baseUrl: this._serverMetadata.openid4vci_metadata.authorization_endpoint,
-      uriTypeProperties: ['redirect_uri', 'scope'],
+      uriTypeProperties: ['redirect_uri', 'scope', 'authorization_details']
     });
 
     return authRequestUrl;
   }
 
+  private addLocations(authorizationDetails?: AuthorizationDetails | AuthorizationDetails[]): void {
+    if (authorizationDetails && this.serverMetadata.openid4vci_metadata?.authorization_server) {
```
nklomp, line 112: Have you looked at the other property pointing to an AS which we are seeing in the wild?

packages/client/lib/OpenID4VCIClient.ts
```diff
@@ -71,37 +71,55 @@ export class OpenID4VCIClient {
...
     }
 
+    this.addLocations(authorizationDetails );
+
     const queryObj: AuthorizationRequest = {
       response_type: ResponseType.AUTH_CODE,
       client_id: clientId,
       code_challenge_method: codeChallengeMethod,
       code_challenge: codeChallenge,
+      authorization_details: JSON.stringify(authorizationDetails),
```
nklomp, line 98: Hmzz, this probably should be encoded right?

packages/common/lib/types/Authorization.types.ts
```diff
@@ -1,5 +1,14 @@
 import { IssuanceInitiationRequestPayload, IssuanceInitiationWithBaseUrl } from './CredentialIssuance.types';
 import { EndpointMetadata, ErrorResponse, PRE_AUTH_CODE_LITERAL } from './Generic.types';
+import {CredentialFormat} from "@sphereon/ssi-types";
+
+export interface AuthorizationDetails {
+  type: 'openid_credential' | string
+  // If the Credential Issuer metadata contains an authorization_server parameter, the authorization detail's locations common data field MUST be set to the Credential Issuer Identifier value.
+  locations?: string | string[]
```
nklomp, line 8: Since this is a property exposed to others, only use the array, as that is the only example we have right now. (Do accept a singular string on the client side though)

packages/client/lib/OpenID4VCIClient.ts
```diff
@@ -71,37 +71,55 @@ export class OpenID4VCIClient {
...
     });
 
     return authRequestUrl;
   }
 
+  private addLocations(authorizationDetails?: AuthorizationDetails | AuthorizationDetails[]): void {
+    if (authorizationDetails && this.serverMetadata.openid4vci_metadata?.authorization_server) {
+      if (Array.isArray(authorizationDetails)) {
+        authorizationDetails.forEach((value, index, array) => {
+          array[index].locations = this.serverMetadata.issuer
```
nklomp, line 115: We should be able to add other locations to

nklomp (changes requested):
See remarks

packages/client/lib/OpenID4VCIClient.ts
maikel-maas: According to the comment, authorization_server is the property used in the wild and the specs are very specific to use it here.

packages/client/lib/OpenID4VCIClient.ts
maikel-maas: It gets encoded at line 110

packages/common/lib/types/Authorization.types.ts
maikel-maas: So, do you want me to create another interface to be accepted as input on the client side or it should be a "DTO" to shape the output?

packages/client/lib/OpenID4VCIClient.ts
nklomp: Okay, so why are you then only using the one which is not spec compliant here?

```` typescript
authorization_server?: string; //NON-SPEC compliant, but used by some issuers. URL of the AS. This URL MUST use the https scheme and MAY contain port, path and query parameter components.
authorization_endpoint?: string;````

File: packages/client/lib/OpenID4VCIClient.ts (hunk 4 above)
Comment (maikel-maas) [reply packages/client/lib/OpenID4VCIClient.ts]: Fixed

File: packages/client/lib/OpenID4VCIClient.ts (hunk 1 above)
Comment (maikel-maas) [reply packages/client/lib/OpenID4VCIClient.ts:119]: According to the [spec](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-5.1.1-5): "If the Credential Issuer metadata contains an authorization_server parameter, the authorization detail's locations common data field MUST be set to the Credential Issuer Identifier value".

File: packages/client/lib/OpenID4VCIClient.ts (hunk 4 above)
Comment (maikel-maas) [reply packages/client/lib/OpenID4VCIClient.ts]: Fixed!

File: packages/client/lib/OpenID4VCIClient.ts (hunk 1 above)
Comment (nklomp) [reply packages/client/lib/OpenID4VCIClient.ts]: Okay than it probably changed in later specs. Still we have both values in the wild. Meaning we need to accomodate for both

File: packages/client/lib/OpenID4VCIClient.ts (hunk 2 above)
Comment (nklomp) [reply packages/client/lib/OpenID4VCIClient.ts]: That is for the whole object. Have you checked whether the invidivual param value should not already be encoded to begin with. Most oAuth2 and OpenID4SSI specs are mentioning that on specific params

**Review by nklomp — CHANGES_REQUESTED** [review]
See remarks about encoding and auth endpoints

File: packages/client/lib/OpenID4VCIClient.ts (hunk 2 above)
Comment (maikel-maas) [reply packages/client/lib/OpenID4VCIClient.ts]: The `authorization_details` property is an object with 3 defined properties: `type` (mandatory), `format` (mandatory) and `locations` (optional). The examples provide additional properties, but they don't show up consistently among the examples and they don't show up anywhere else in the documentation. The examples below show the json string encoded:

[OpenID4VCI](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-5.1.1-7)
```
GET /authorize?
  response_type=code
  &client_id=s6BhdRkqt3
  &code_challenge=E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM
  &code_challenge_method=S256
  &authorization_details=%5B%7B%22type%22:%22openid_credential
  %22,%22format%22:%22jwt_vc_json%22,%22types%22:%5B%22Verifia
  bleCredential%22,%22UniversityDegreeCredential%22%5D%7D%5D
  &redirect_uri=https%3A%2F%2Fclient.example.org%2Fcb
Host: https://server.example.com
``` 

[draft-ietf-oauth-rar-23 ](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-rar-23#name-authorization-request)
GET /authorize?response_type=code
   &client_id=s6BhdRkqt3
   &state=af0ifjsldkj
   &redirect_uri=https%3A%2F%2Fclient.example.org%2Fcb
   &code_challenge_method=S256
   &code_challenge=K2-ltc83acc4h0c9w6ESC_rEMTJ3bwc-uCHaoeK1t8U
   &authorization_details=%5B%7B%22type%22%3A%22account%5Finfo
   rmation%22%2C%22actions%22%3A%5B%22list%5Faccounts%22%2C%22
   read%5Fbalances%22%2C%22read%5Ftransactions%22%5D%2C%22loca
   tions%22%3A%5B%22https%3A%2F%2Fexample%2Ecom%2Faccounts%22%
   5D%7D%2C%7B%22type%22%3A%22payment%5Finitiation%22%2C%22act
   ions%22%3A%5B%22initiate%22%2C%22status%22%2C%22cancel%22%5
   D%2C%22locations%22%3A%5B%22https%3A%2F%2Fexample%2Ecom%2Fp
   ayments%22%5D%2C%22instructedAmount%22%3A%7B%22currency%22%
   3A%22EUR%22%2C%22amount%22%3A%22123%2E50%22%7D%2C%22credito
   rName%22%3A%22Merchant%20A%22%2C%22creditorAccount%22%3A%7B
   %22iban%22%3A%22DE02100100109307118603%22%7D%2C%22remittanc
   eInformationUnstructured%22%3A%22Ref%20Number%20Merchant%22
   %7D%5D HTTP/1.1
Host: server.example.com

Both show everything encoded

File: packages/client/lib/OpenID4VCIClient.ts (hunk 1 above)
Comment (maikel-maas) [reply packages/client/lib/OpenID4VCIClient.ts]: Fixed!

File: packages/client/lib/AuthorizationDetailsBuilder.ts [hunk 5]
```diff
@@ -0,0 +1,36 @@
...
+    this.authorizationDetails.type = type;
+    return this;
+  }
+
+  withFormats(format: CredentialFormat): AuthorizationDetailsBuilder {
+    this.authorizationDetails.format = format;
+    return this;
+  }
+
+  withLocations(locations: string[]): AuthorizationDetailsBuilder {
```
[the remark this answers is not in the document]
nklomp: Not what I asked!

nklomp (approved):
LGTM

PR #34, opened 2023-03-28
Feature/vdx 172/version discovery on develop

hrehman-sphereon:


packages/client/lib/AccessTokenClient/CredentialOfferAccessTokenClient.ts
```diff
@@ -0,0 +1,53 @@
...
+  IssuerOpts,
+  OpenIDResponse,
+} from '@sphereon/openid4vci-common';
+
+import { IssuanceInitiationClient } from '../CredentialOffer';
+
+import { AccessTokenClient } from './AccessTokenClient';
+
+// noinspection JSUnusedLocalSymbols
+export class CredentialOfferAccessTokenClient implements AccessTokenClient {
```
nklomp, line 17: Why is this whole class needed according to you at this point? Where in the spec does V11 warrant a new implementation for the AccessTokenClient next to v9?

The more classes we create, the more maintenance and potential for bugs we introduce.

Also I rather see a suffix denoting the Version of the spec if we need seperate classes. The access token client itself has almost zero todo with Credential offers or Initiate Issuance. It is litterally an OAuth2 AccessToken client. By introducing these names it becomes confusing.

packages/client/lib/AccessTokenClient/IssuanceInitiationAccessTokenClient.ts
```diff
@@ -13,38 +13,55 @@ import {
...
+    issuanceInitiationClient: IssuanceInitiationClient,
+    metadata: EndpointMetadata,
+    pin?: string,
+    clientId?: string,
+    codeVerifier?: string,
+    code?: string,
+    redirectUri?: string
+  ): IssuanceInitiationAccessTokenRequestOpts => {
+    return {
+      issuanceInitiation: issuanceInitiationClient.issuanceInitiationWithBaseUrl,
```
nklomp, line 35: Please just stick to the assignment at hand. The previous destructured aproach was more legable backed by an interface

packages/client/lib/AccessTokenClient/IssuanceInitiationAccessTokenClient.ts
```diff
@@ -13,38 +13,55 @@ import {
...
+      metadata: metadata,
```
nklomp, line 36: No need to repeat. `metadata` suffices as the json will automatically now have the key 'metdata' with the values from the metadata object

packages/client/lib/AccessTokenClient/IssuanceInitiationAccessTokenClient.ts
```diff
@@ -13,38 +13,55 @@ import {
...
+      pin,
+      codeVerifier,
+      code,
+      redirectUri,
+      asOpts: { clientId: clientId },
```
nklomp, line 41: No repetition necessary

packages/client/lib/AccessTokenClient/IssuanceInitiationAccessTokenClient.ts
```diff
@@ -13,38 +13,55 @@ import {
...
 
     return await this.acquireAccessTokenUsingRequest({
       accessTokenRequest: await this.createAccessTokenRequest({
-        issuanceInitiation,
-        asOpts,
-        codeVerifier,
-        code,
-        redirectUri,
-        pin,
+        issuanceInitiation: issuanceInitiation,
```
nklomp, line 55: Please just stick to the assignment at hand. The previous was more legable  as it used destructuring, so you do not get cluttered types that do not add much, as you are basically taking the inputs and using them again as arguments for a next function

packages/client/lib/AccessTokenClient/IssuanceInitiationAccessTokenClient.ts
```diff
@@ -13,38 +13,55 @@ import {
...
+      code,
+      redirectUri,
+      asOpts: { clientId: clientId },
+    };
+  };
 
-    const isPinRequired = this.isPinRequiredValue(issuanceInitiationRequest);
-    const issuerOpts = { issuer: issuanceInitiationRequest.issuer };
+  public async acquireAccessToken(accessTokenRequestOpts: IssuanceInitiationAccessTokenRequestOpts): Promise<OpenIDResponse<AccessTokenResponse>> {
+    const accessTokenRequestOptions = accessTokenRequestOpts as IssuanceInitiationAccessTokenRequestOpts;
```
nklomp, line 46: Why is this cast necessary? Also why is there a need for a copy when the param is just being used to read the options?

packages/common/lib/types/OpenID4VCIVersions.types.ts
```diff
@@ -0,0 +1,10 @@
+export enum OpenId4VCIVersion {
+  VER_9 = 'version9',
```
nklomp, line 2: Please assign ints, so we can do things like `if (ver > VER_9.valueOf())`

packages/common/lib/types/OpenID4VCIVersions.types.ts
```diff
@@ -0,0 +1,10 @@
...
+  VER_11 = 'version11',
+  VER_UNKNOWN = 'unknown version',
+}
+
+export enum URLSchemes {
```
nklomp, line 7: These are URI Schemes. Suggest to rename to DefaultURISchemes, as there always is the option to use your own scheme

packages/client/lib/__tests__/OpenID4VCIClient.spec.ts
```diff
@@ -10,8 +10,8 @@ describe('OpenID4VCIClient should', () => {
 
   beforeEach(async () => {
     nock(MOCK_URL).get(/.*/).reply(200, {});
-    client = await OpenID4VCIClient.initiateFromURI({
-      issuanceInitiationURI: 'openid-initiate-issuance://?issuer=https://server.example.com&credential_type=TestCredential',
+    client = await OpenID4VCIClient.fromURI({
+      credentialOfferURI: 'openid-initiate-issuance://?issuer=https://server.example.com&credential_type=TestCredential',
```
nklomp, line 14: 2 things:

1) The param name doesn't match the actual URI. This of course has to do with the rename in V11. Credential Offer is not a concept in V9, which then could confuse people.

2) The param should simply be renamed to uri for both versions, similar like the lower level class working with URIs. Previously we didn't know the actual concept of initiation would be completely renamed in the future.  The difference should become apparent from the actual value (openid-initiate-issuance vs openid-credential-offer). Since we now have 2 different URI scheme's and in the future you should also be able to have custom scheme's, simply renaming it to uri makes sense (similar to the method name)

packages/client/lib/__tests__/IT.spec.ts
```diff
@@ -30,10 +37,13 @@ describe('OID4VCI-Client should', () => {
...
-  const INITIATE_QR_DATA =
+  const INITIATE_QR_DATA_VER9 =
     'openid-initiate-issuance://?issuer=https%3A%2F%2Fissuer.research.identiproof.io&credential_type=OpenBadgeCredentialUrl&pre-authorized_code=4jLs9xZHEfqcoow0kHE7d1a8hUk6Sy-5bVSV2MqBUGUgiFFQi-ImL62T-FmLIo8hKA1UdMPH0lM1xAgcFkJfxIw9L-lI3mVs0hRT8YVwsEM1ma6N3wzuCdwtMU4bcwKp&user_pin_required=true';
 
-  it('succeed with a full flow wit the client', async () => {
+  const INITIATE_QR_DATA_VER11 =
+    'openid-credential-offer://credential_offer=%7B%22credential_issuer%22:%22https://credential-issuer.example.com%22,%22credentials%22:%5B%7B%22format%22:%22jwt_vc_json%22,%22types%22:%5B%22VerifiableCredential%22,%22UniversityDegreeCredential%22%5D%7D%5D,%22issuer_state%22:%22eyJhbGciOiJSU0Et...FYUaBy%22%7D';
```
nklomp, line 44: OFFER_QR or CREDENTIAL_OFFER_QR for v11

packages/client/lib/MetadataClient/IssuanceInitiationMetadataClient.ts
```diff
@@ -0,0 +1,22 @@
+import { EndpointMetadata, IssuanceInitiationRequestPayload, IssuanceInitiationWithBaseUrl } from '@sphereon/openid4vci-common';
```
nklomp, line 1: Is there a good reason to split this up already? Why not simply accept both versions? It only is a name change at this point.

packages/client/lib/MetadataClient/CredentialOfferMetadataClient.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { MetadataClient } from './MetadataClient';
+
+export class CredentialOfferMetadataClient extends MetadataClient {
+  /**
+   * Retrieve metadata using the credentialOfferWithBaseURL obtained from a previous step
+   *
+   * @param credentialOfferWithBaseURL
+   */
+  public static async getServerMetaData(credentialOfferWithBaseURL: CredentialOfferWithBaseURL): Promise<EndpointMetadata> {
```
nklomp, line 11: Is there a good reason to split this up already? Why not simply accept both versions? It only is a name change at this point.

packages/client/lib/CredentialOffer/CredentialOfferClient.ts
```diff
@@ -0,0 +1,42 @@
+import { CredentialOfferWithBaseURL, OpenId4VCIVersion } from '@sphereon/openid4vci-common';
+import Debug from 'debug';
+
+import { CredentialIssuanceClient } from './CredentialIssuanceClient';
+
+const debug = Debug('sphereon:openid4vci:credentialOffer');
+
+// FIXME implement the function
+export class CredentialOfferClient implements CredentialIssuanceClient {
```
nklomp, line 9: Why is a new client necessary? The change is a simple change in naming/URI. The more classes we introduce the more maintenance and the more chances for bugs.

nklomp (changes requested):
You seem to be changing way to much. Just stick to the assignment. 

The more classes we introduce the more maintenance and the more room for bugs. The biggest change is the URI schema name change and then the different payload names, which gets apparent in some properties. 

Unless I am missing something here, there really is not need to start introducing all kinds of V11 versions for classes being used later by the client. Simply make it uniform early on by applying the version discovery on the URI, then adjusting some payload potentially. 

I don't see what all these classes bring except for complexity versus simply accepting a union of arguments in maybe 3 or 4 places.

packages/client/lib/AccessTokenClient/IssuanceInitiationAccessTokenClient.ts
hrehman-sphereon: fixed

packages/client/lib/AccessTokenClient/IssuanceInitiationAccessTokenClient.ts
hrehman-sphereon: fixed

packages/client/lib/AccessTokenClient/IssuanceInitiationAccessTokenClient.ts
hrehman-sphereon: fixed

packages/common/lib/types/OpenID4VCIVersions.types.ts
hrehman-sphereon: fixed

packages/client/lib/AccessTokenClient/CredentialOfferAccessTokenClient.ts
hrehman-sphereon: All big things start small. And based on OO and design patterns this was supposed to be separate. This was also discussed on call with you and this approach was approved. Once the version 9 needs to be removed we will not need to surgically remove the version 9 specific details from single file.

I agree, it should have received only the data that the class needed rather than "initiation" object so that it should not have come to this.

Actually, I disagree with using union types ... As tomorrow for next version will we add another union type through out the code again ? 

@nklomp Shall I rename the files instead of removing these ?

packages/client/lib/AccessTokenClient/IssuanceInitiationAccessTokenClient.ts
hrehman-sphereon: fixed

packages/client/lib/AccessTokenClient/CredentialOfferAccessTokenClient.ts
hrehman-sphereon: @nklomp Reverted.

packages/client/lib/AccessTokenClient/IssuanceInitiationAccessTokenClient.ts
hrehman-sphereon: reverted

packages/client/lib/MetadataClient/CredentialOfferMetadataClient.ts
hrehman-sphereon: reverted.

packages/client/lib/MetadataClient/IssuanceInitiationMetadataClient.ts
hrehman-sphereon: reverted

packages/client/lib/CredentialOffer/CredentialOfferClient.ts
hrehman-sphereon: reverted.

nklomp (changes requested):
Some small remarks. Overall going in the correct direction

packages/client/lib/OpenID4VCIClient.ts
```diff
@@ -305,8 +288,10 @@ export class OpenID4VCIClient {
     const credentialsSupported = this.serverMetadata?.openid4vci_metadata?.credentials_supported;
     if (!credentialsSupported) {
       return {};
-    } else if (!restrictToInitiationTypes === false) {
-        return credentialsSupported;
+    } else { // noinspection PointlessBooleanExpressionJS
+      if (restrictToInitiationTypes === false) {
```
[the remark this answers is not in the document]
hrehman-sphereon: Intellij tired auto correct ... reverted.

PR #36, opened 2023-03-30
feature/VDX-178/pushed_auth_request

maikel-maas:


package.json
```diff
@@ -6,6 +6,7 @@
   "license": "Apache-2.0",
   "private": true,
   "scripts": {
+    "start": "ts-node packages/issuer/lib/AuthServer.ts",
```
nklomp, line 9: Please move that to the actual module. The REST API for instance will move to another module. So make it the concern of these modules themselves

packages/client/lib/OpenID4VCIClient.ts
```diff
@@ -136,7 +137,46 @@ export class OpenID4VCIClient {
...
 
-  private handleAuthorizationDetails(authorizationDetails?: AuthDetails | AuthDetails[]): AuthDetails | AuthDetails[] | undefined {
+  public async acquirePushedAuthorizationRequestURI({
+    clientId,
+    codeChallengeMethod,
+    codeChallenge,
+    authorizationDetails,
+    redirectUri,
+    scope,
+  }: AuthRequestOpts): Promise<OpenIDResponse<unknown>> {
```
nklomp, line 147: Why the unknown? You know what payload type to expect right?

packages/common/lib/types/Generic.types.ts
```diff
@@ -202,5 +202,7 @@ export interface EndpointMetadata {
   token_endpoint: string;
   credential_endpoint: string;
   authorization_endpoint?: string;
+  pushed_authorization_request_endpoint?: string;
+  require_pushed_authorization_endpoint?: boolean;
```
nklomp, line 206: I think I asked this already, but which spec does this value come from? This is a payload. We cannot makeup something ourselves here (if that is the case)

Didn't you happen to mean require_pushed_authorization_requests from https://www.rfc-editor.org/rfc/rfc9126#section-6

packages/common/lib/types/OpenID4VCIServerMetadata.ts
```diff
@@ -12,6 +12,9 @@ export interface OpenID4VCIServerMetadata {
...
   // TODO: The above authorization_server being used in the wild, serves roughly the same purpose as the below spec compliant endpoint. Look at how to use authorization_server as authorization_endpoint in case it is present
   authorization_endpoint?: string;
+  pushed_authorization_request_endpoint?: string; // The URL of the pushed authorization request endpoint at which a client can post an authorization request to exchange for a request_uri value usable at the authorization server
+  // Note that the presence of pushed_authorization_request_endpoint is sufficient for a client to determine that it may use the PAR flow. A request_uri value obtained from the PAR endpoint is usable at the authorization endpoint regardless of other authorization server metadata such as request_uri_parameter_supported or require_request_uri_registration
+  require_pushed_authorization_endpoint?: boolean; // Boolean parameter indicating whether the authorization server accepts authorization request data only via PAR. If omitted, the default value is false.
```
nklomp, line 17: I think I asked this already, but which spec does this value come from? This is a payload. We cannot makeup something ourselves here (if that is the case)
Didn't you happen to mean require_pushed_authorization_requests from https://www.rfc-editor.org/rfc/rfc9126#section-5 ?

packages/issuer-rest/lib/api.ts
```diff
@@ -97,37 +99,82 @@ export class RestAPI {
...
+        return res.status(400).json({
+          error: 'invalid_request',
+          error_description: message,
+        })
+      }
+      return next()
+    }
+
+    this.express.post('/par', handleHttpStatus400, (req: Request, res: Response) => {
+      // Fake client for testing, it needs to come from a registered client
```
nklomp, line 145: Please make this a fixme and create a ticket for it

packages/issuer-rest/lib/expressUtils.ts
```diff
@@ -0,0 +1,21 @@
+import { Request } from 'express'
+
+export const validateRequestBody = ({
```
nklomp, line 3: Either make the name more clear, or just throw an error from here (and catch it).

nklomp (changes requested):
See some initial remarks. Need to do a more thorough review later

packages/common/lib/types/Generic.types.ts
maikel-maas: It's from [here](https://www.rfc-editor.org/rfc/rfc9126#name-authorization-server-metada) actually and it indicates whether the authorization server accepts authorization request data only via PAR.

packages/common/lib/types/OpenID4VCIServerMetadata.ts
maikel-maas: Yes

package.json
maikel-maas: Fixed!

packages/client/lib/OpenID4VCIClient.ts
maikel-maas: Fixed!

packages/common/lib/types/Generic.types.ts
maikel-maas: It's from [section 5](https://www.rfc-editor.org/rfc/rfc9126#name-authorization-server-metada). It indicates whether the authorization server accepts authorization request data only via PAR

packages/issuer-rest/lib/api.ts
maikel-maas: Fixed!

packages/issuer-rest/lib/expressUtils.ts
maikel-maas: Fixed

packages/issuer-rest/lib/api.ts
```diff
@@ -97,37 +99,83 @@ export class RestAPI {
...
+
+      // The scopes from the request need to be matched against the ones present in the registered client
+      if (!req.body.scope.split(',').every((scope: string) => client.scope.includes(scope))) {
+        return res.status(400).send({ error: 'invalid_scope', error_description: 'scope is not valid for the given client' })
+      }
+
+      //TODO Implement authorization_details verification
+
+      const uuid = uuidv4()
+      const requestUri = `urn:ietf:params:oauth:request_uri:${uuid}`
```
nklomp, line 166: The base URL for sure would also need to be optionally provided by the developer. Meaning there should be a configuration option fo the base URI (so everything in front of the uuid). This also needs a TODO, as in the future you even might want to bind the request uri to a specific client for instance

nklomp (changes requested):
See remarks

packages/common/lib/types/Generic.types.ts
nklomp: Values should not be in this object to begin with. This is an object we create based on retrieval of several metadata endpoints. The value should be in the OAuth2 Server metadata, so also not in the IssuerMetadata. Will fix that

packages/common/lib/types/OpenID4VCIServerMetadata.ts
nklomp: Values should not be in this object to begin with. This is an object we create based on retrieval of several metadata endpoints. The value should be in the OAuth2 Server metadata, so also not in the IssuerMetadata. Will fix that

nklomp (approved):

PR #39, opened 2023-04-03
Feature/vdx 171 issuer module

sksadjad:
issuer related builder plus some changes and api.ts class added.


packages/client/lib/AuthorizationDetailsBuilder.ts
```diff
@@ -36,9 +35,9 @@ export class AuthorizationDetailsBuilder {
     return this;
   }
 
-  build(): AuthorizationDetails {
+  build(): CommonAuthorizationDetails {
```
nklomp, line 38: Hmzz, I also would like to see specific types being build. Otherwise the user will have to cast certain cases

packages/common/lib/types/Authorization.types.ts
```diff
@@ -1,14 +1,51 @@
...
 
-export interface AuthorizationDetails {
+export interface CommonAuthorizationDetails {
   type: 'openid_credential' | string;
+  format: CredentialFormatEnum;
   // If the Credential Issuer metadata contains an authorization_server parameter, the authorization detail's locations common data field MUST be set to the Credential Issuer Identifier value.
   locations?: string[];
-  format: CredentialFormat;
-  [s: string]: unknown;
+  [key: string]: unknown;
```
nklomp, line 36: This is probably one of the few cases any is better in case the values can also be objects. Reason is that you then can access subproperties. On an unknown JSC will throw errors

packages/common/lib/types/Authorization.types.ts
```diff
@@ -68,7 +105,7 @@ export interface AuthorizationRequestOpts {
   clientId: string;
   codeChallenge: string;
   codeChallengeMethod: CodeChallengeMethod;
-  authorizationDetails?: AuthorizationDetails[];
+  authorizationDetails?: CommonAuthorizationDetails[];
```
nklomp, line 108: I rather see a union here. Problem is that I rather force a developer to accomodate for multiple options, but knowing which properties are available, then having a super interface, and a developer missing certain properties that are present in the subclass/interface

packages/common/lib/types/CredentialIssuance.types.ts
```diff
@@ -1,11 +1,11 @@
...
-import {OpenId4VCIVersion} from "./OpenID4VCIVersions.types";
+
+import { OpenId4VCIVersion } from './OpenID4VCIVersions.types';
 
 export interface CredentialRequest {
   //TODO: handling list is out of scope for now
   type: string | string[];
-  //TODO: handling list is out of scope for now
-  format: CredentialFormat | CredentialFormat[];
+  format: CredentialFormat;
```
nklomp, line 8: Have you doublechecked this? I rather have us conform whenever we send out things, but be lenient when we get things

packages/common/lib/types/Generic.types.ts
```diff
@@ -36,7 +35,10 @@ export interface IssuerMetadata {
   credential_endpoint: string;
   batch_credential_endpoint?: string;
   credentials_supported: CredentialIssuerMetadataSupportedCredentials[];
-  credential_issuer: Display;
+  credential_issuer: string;
+  authorization_server?: string;
+  token_endpoint?: string;
+  display?: (NameAndLocale & { [key: string]: string })[];
```
nklomp, line 41: Would be nice to make a type or interface out of this

packages/issuer/lib/VcIssuer.ts
```diff
@@ -0,0 +1,56 @@
...
+          if (credentialSupported.format === format) {
+            return true
+          }
+        }
+      }
+    }
+    return false
+  }
+
+  private async issueCredential(issueCredentialRequest: CredentialRequest): Promise<CredentialResponse> {
```
nklomp, line 39: Just create a callback function that accepts both the CredentialRequest as well as a W3CCredential/ICredential

packages/issuer/lib/api.ts
```diff
@@ -0,0 +1,137 @@
...
+import express, { Express, Response } from 'express'
+import { v4 as uuidv4 } from 'uuid'
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+
+import { VcIssuer } from './VcIssuer'
+import { CredentialSupportedV1_11Builder, VcIssuerBuilder } from './builder'
+import { createCredentialOfferDeeplink } from './functions'
+
+function buildVcIssuer() {
```
nklomp, line 20: buildVCIFromEnvironment, as we certainly want to have other means to instantiate issuers in the future as well

packages/issuer/lib/builder/VcIssuerBuilder.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.batchCredentialEndpoint = batchCredentialEndpoint
+    return this
+  }
+
+  public withTokenEndpoint(tokenEndpoint: string): VcIssuerBuilder {
+    this.tokenEndpoint = tokenEndpoint
+    return this
+  }
+
+  public withIssuerDisplay(issuerDisplay: Display[] | Display): VcIssuerBuilder {
```
nklomp, line 39: with doesn 't suggest merging with existing property values. With suggests overwriting them.

Suggest to have a withXX which overwrites (IssuerDisplay and CredentialsSupported) and a seperate addXX to add a singular value for (IssuerDisplay and CredentialSupported)

packages/issuer/lib/functions/CredentialOffer.ts
```diff
@@ -0,0 +1,35 @@
+import {
+  CredentialFormatEnum,
+  CredentialIssuerMetadataSupportedCredentials,
+  encodeJsonAsURI,
+  IssuerMetadata,
+  TokenErrorResponse,
+} from '@sphereon/openid4vci-common'
+import { v4 as uuidv4 } from 'uuid'
+
+export function createCredentialOfferDeeplink(preAuthorizedCode: string, issuerMetadata: IssuerMetadata): string {
```
nklomp, line 10: createCredentialOfferURI

packages/issuer/lib/api.ts
```diff
@@ -0,0 +1,137 @@
...
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    this.express.post('/token', (request, response) => {
+      return RestAPI.sendErrorResponse(response, 501, 'Not implemented')
+    })
+  }
+  private registerCredentialRequestEndpoint() {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    this.express.get('/credential-request', async (request, response) => {
```
nklomp, line 120: Why is this a GET? Doesn't our client lib POST to the credentialOffer endpoint and doesn't the spec say POST?https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-7.2

nklomp (changes requested):
See remarks

packages/issuer/lib/api.ts
sksadjad: changed this to `post`

packages/issuer/lib/api.ts
sksadjad: changed this method to `buildVCIFromEnvironment` and also added another option to build `api` object with an existing metadata object

packages/issuer/lib/builder/VcIssuerBuilder.ts
sksadjad: changed this method and several other places that we have a similar pattern

packages/common/lib/types/Authorization.types.ts
sksadjad: changed to `any`

packages/common/lib/types/Authorization.types.ts
sksadjad: added a union type for `AuthorizationDetails`.

packages/common/lib/types/CredentialIssuance.types.ts
sksadjad: it should be singular, but I reverted this for conformity

packages/issuer/lib/VcIssuer.ts
sksadjad: added callback function

packages/common/lib/types/v1_0_09.types.ts
```diff
@@ -15,19 +16,10 @@ export interface AuthorizationRequestV1_0_09 extends CommonAuthorizationRequest
...
-  [key: string]: unknown;
-}
-
 export function isAuthorizationRequestV1_0_09(request: CommonAuthorizationRequest): boolean {
   return request && 'op_state' in request;
 }
 
-export function isCredentialOfferV1_0_09(request: CredentialOfferV1_0_09 | CredentialOfferV1_0_11): boolean {
-  return request && 'issuer' in request && 'op_state' in request;
+export function isCredentialOfferV1_0_09(request: CredentialOfferPayload): boolean {
```
nklomp, line 23: Probably wise to create an isCredentialOfferVersion(request, version) method as a TODO, for these, otherwise we end up with 10 of them in the future.

Also having something like isCredentialOfferAtLeast or something makes sense for the future

nklomp (changes requested):
See remarks,

I am a little bit worried too much is gone. If I simply look in Github the amount of interfaces/types with comments now gone is pretty big.

Also I fail to see why we already are calling the builder a certain version, whilst it doesn't really do something specific for versions. In other words, probably we can stick with a single builder, building for different versions for now. It is mainly some name changes and small payload changes at this point.

Lastly add fixmes pointing to a ticket in places where you are now casting to V9 in the client, so we at least know which points to touch when starting work on v11 for the client.

packages/client/lib/CredentialOffer.ts
```diff
@@ -20,15 +20,15 @@ export class CredentialOffer {
...
-    }) as IssuanceInitiationRequestPayloadV9;
+    }) as CredentialOfferV1_0_09;
 
     const request =
-        version < OpenId4VCIVersion.VER_11.valueOf()
-            ? (issuanceInitiationRequest as IssuanceInitiationRequestPayloadV9)
-            : (issuanceInitiationRequest as CredentialOfferRequestPayloadV11);
+      version < OpenId4VCIVersion.VER_1_0_11.valueOf()
+        ? (issuanceInitiationRequest as CredentialOfferV1_0_09)
+        : (issuanceInitiationRequest as unknown as CredentialOfferPayloadV1_0_11);
```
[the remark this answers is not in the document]
sksadjad: yeah it was, thanks for noticing. the original code was just looking for properties of v9, so it couldn't generate v11 there. added a few lines of code for casting to v11

packages/client/lib/CredentialRequestV1_0_09ClientBuilder.ts
```diff
@@ -1,28 +1,29 @@
...
+export class CredentialRequestV1_0_09ClientBuilder {
   credentialEndpoint?: string;
   credentialType?: string | string[];
   format?: CredentialFormat | CredentialFormat[];
   token?: string;
 
-  public static fromURI({ uri, metadata }: { uri: string; metadata?: EndpointMetadata }): CredentialRequestClientBuilder {
-    return CredentialRequestClientBuilder.fromCredentialOfferRequest({
+  public static fromURI({ uri, metadata }: { uri: string; metadata?: EndpointMetadata }): CredentialRequestV1_0_09ClientBuilder {
+    return CredentialRequestV1_0_09ClientBuilder.fromCredentialOfferRequest({
```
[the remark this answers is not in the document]
sksadjad: we can have a builder with two different methods. definitely. but it will make a messy builder. because for example while in the [v9](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-09.html#name-issuance-initiation-request) we have `pre-authorized_code` and `user_pin_required`we have these in the grants in [v11](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-11.html#name-credential-offer-parameters). the other properties are slightly different as well. so for me it makes more sense to isolate v9 codes here. because continuing the support for it in the main code will cause a lot of confusion later

packages/client/lib/AccessTokenClient.ts
```diff
@@ -111,7 +115,7 @@ export class AccessTokenClient {
       request.redirect_uri = redirectUri;
       request.grant_type = GrantTypes.AUTHORIZATION_CODE;
     }
-    if (request.grant_type === GrantTypes.AUTHORIZATION_CODE && credentialOfferRequest[PRE_AUTH_CODE_LITERAL]) {
+    if (request.grant_type === GrantTypes.AUTHORIZATION_CODE && (credentialOfferRequest as CredentialOfferV1_0_09)[PRE_AUTH_CODE_LITERAL]) {
```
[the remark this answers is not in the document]
sksadjad: added todo for v11

packages/client/lib/AccessTokenClient.ts
```diff
@@ -94,12 +98,12 @@ export class AccessTokenClient {
...
     request.user_pin = pin;
 
-    if (credentialOfferRequest[PRE_AUTH_CODE_LITERAL]) {
+    if (credentialOfferRequest[PRE_AUTH_CODE_LITERAL as keyof CredentialOfferPayload]) {
       if (codeVerifier) {
         throw new Error('Cannot pass a code_verifier when flow type is pre-authorized');
       }
       request.grant_type = GrantTypes.PRE_AUTHORIZED_CODE;
-      request[PRE_AUTH_CODE_LITERAL] = credentialOfferRequest[PRE_AUTH_CODE_LITERAL];
+      request[PRE_AUTH_CODE_LITERAL] = (credentialOfferRequest as CredentialOfferV1_0_09)[PRE_AUTH_CODE_LITERAL];
```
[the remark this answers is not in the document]
sksadjad: added todo for v11

packages/client/lib/AccessTokenClient.ts
```diff
@@ -19,21 +25,19 @@ import { convertJsonToURI, formPost } from './functions';
...
+    codeVerifier,
+    code,
+    redirectUri,
+    metadata,
+  }: AccessTokenRequestOpts): Promise<OpenIDResponse<AccessTokenResponse>> {
     const { request } = credentialOffer;
 
     const isPinRequired = this.isPinRequiredValue(request);
-    const issuerOpts = { issuer: request.issuer };
+    const issuerOpts = { issuer: 'issuer' in request ? request.issuer : 'credential_issuer' in request ? request.credential_issuer : '' };
```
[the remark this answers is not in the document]
sksadjad: changed to use the utils method for this, which thrown an exception.

packages/client/lib/OpenID4VCIClient.ts
```diff
@@ -118,15 +119,15 @@ export class OpenID4VCIClient {
...
+    const queryObj = {
       response_type: ResponseType.AUTH_CODE,
       client_id: clientId,
       code_challenge_method: codeChallengeMethod,
       code_challenge: codeChallenge,
       authorization_details: JSON.stringify(this.handleAuthorizationDetails(authorizationDetails)),
       redirect_uri: redirectUri,
       scope: scope,
-    };
+    } as AuthorizationRequestV1_0_09;
```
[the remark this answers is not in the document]
sksadjad: added fixme for v11

packages/client/lib/AccessTokenClient.ts
```diff
@@ -130,16 +136,26 @@ export class AccessTokenClient {
...
+    if (isCredentialOfferV1_0_09(requestPayload)) {
+      requestPayload = requestPayload as CredentialOfferV1_0_09;
       if (typeof requestPayload.user_pin_required === 'string') {
         isPinRequired = requestPayload.user_pin_required.toLowerCase() === 'true';
       } else if (typeof requestPayload.user_pin_required === 'boolean') {
         isPinRequired = requestPayload.user_pin_required;
       }
+    } else if (isCredentialOfferV1_0_11(requestPayload)) {
+      requestPayload = requestPayload as CredentialOfferPayloadV1_0_11;
+      if ('grants' in requestPayload && 'urn:ietf:params:oauth:grant-type:pre-authorized_code' in requestPayload.grants!) {
```
maikel-maas, line 154: // eslint-disable-next-line @typescript-eslint/no-non-null-assertion

packages/client/lib/AccessTokenClient.ts
```diff
@@ -130,16 +136,26 @@ export class AccessTokenClient {
...
+        isPinRequired = requestPayload!.grants!['urn:ietf:params:oauth:grant-type:pre-authorized_code']!.user_pin_required;
```
maikel-maas, line 155: // eslint-disable-next-line @typescript-eslint/no-non-null-assertion

packages/common/lib/types/Authorization.types.ts
```diff
@@ -1,14 +1,52 @@
...
-export interface AuthorizationDetails {
+export interface CommonAuthorizationDetails {
   type: 'openid_credential' | string;
+  format: CredentialFormatEnum;
   // If the Credential Issuer metadata contains an authorization_server parameter, the authorization detail's locations common data field MUST be set to the Credential Issuer Identifier value.
   locations?: string[];
-  format: CredentialFormat;
-  [s: string]: unknown;
+  types: string[];
+  [key: string]: any;
```
maikel-maas, line 41: // eslint-disable-next-line  @typescript-eslint/no-explicit-any

maikel-maas (commented):
Please run pnpm fix

packages/client/lib/AccessTokenClient.ts
sksadjad: added the above comment

packages/client/lib/AccessTokenClient.ts
sksadjad: added the above comment

packages/common/lib/types/Authorization.types.ts
sksadjad: added the above comment

packages/common/lib/types/v1_0_09.types.ts
sksadjad: added two tickets for `CredentialOffer` and `AuthorizationRequest`

packages/client/lib/OpenID4VCIClient.ts
```diff
@@ -301,14 +302,14 @@ export class OpenID4VCIClient {
...
-  getCredentialMetadata(type: string): CredentialMetadata {
+  getCredentialMetadata(type: string): IssuerCredentialSubjectDisplay {
     return this.getCredentialsSupported(false)[type];
   }
 
   getCredentialTypes(): string[] {
-    return typeof this.credentialOffer.request.credential_type === 'string'
-        ? [this.credentialOffer.request.credential_type]
-        : this.credentialOffer.request.credential_type;
+    return typeof (this.credentialOffer.request as CredentialOfferV1_0_09).credential_type === 'string'
```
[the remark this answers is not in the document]
sksadjad: added todo for this

packages/common/lib/types/Generic.types.ts
```diff
@@ -24,19 +23,24 @@ export interface NameAndLocale {
...
-export type Display = NameAndLocale & LogoAndColor & { [key: string]: string };
+export type Display = NameAndLocaleExtended & LogoAndColor;
 
 export interface IssuerMetadata {
   credential_endpoint: string;
   batch_credential_endpoint?: string;
   credentials_supported: CredentialIssuerMetadataSupportedCredentials[];
-  credential_issuer: Display;
+  credential_issuer: string;
+  authorization_server?: string;
```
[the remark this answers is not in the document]
nklomp: Hmzz, not sure I like that. This means we are getting 2 locations for something which is exactly the same concept. The client and issuer should always have the same payloads so we do not get into situations where the client is behaving differently from the issuer, simply because the interfaces for the payload differ, or we forget to add certain logic in one place.

packages/common/lib/types/OpenID4VCIServerMetadata.ts
```diff
@@ -92,50 +17,14 @@ export interface OpenID4VCIServerMetadata {
 export type Oauth2ASWithOID4VCIMetadata = OAuth2ASMetadata & OpenID4VCIServerMetadata;
 
 export interface CredentialIssuer {
-  display?: IssuerDisplay; //OPTIONAL. An array of objects, where each object contains display properties of a Credential issuer for a certain language. Below is a non-exhaustive list of valid parameters that MAY be included:
+  display?: NameAndLocaleExtended; //OPTIONAL. An array of objects, where each object contains display properties of a Credential issuer for a certain language. Below is a non-exhaustive list of valid parameters that MAY be included:
```
nklomp, line 20: I find it strange that we now have this interface/type instead of the Display type for all kinds of display properties. It is confusing since the Display type is also there. So either this is incorrect, or suggest to start renaming some.

Something similar happens above for the credentials_supported. The type name is worse than what it was

packages/issuer/lib/VcIssuer.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  public getIssuerMetadata() {
+    return this._issuerMetadata
+  }
+
+  public async issueCredentialFromIssueRequest(issueCredentialRequest: CredentialRequest): Promise<CredentialResponse> {
+    //TODO: do we want additional validations here?
+    if (this.isMetadataSupportCredentialRequestFormat(issueCredentialRequest.format)) {
+      return {
+        credential: await this.issueCredential({ credentialRequest: issueCredentialRequest }),
```
nklomp, line 22: Probably wise to have an issuerCallback as optional argument for this method. If supplied that one is used instead of the callback set at the class level. Sometimes you need more specific context/variables when issuing a credential. Also do not only supply the arguments to the callback function alone. The whole callback function would need to be passed in next to the arguments, so an implementer has full control over how they want to issue

packages/issuer/lib/builder/CredentialSupportedV1_11Builder.ts
```diff
@@ -0,0 +1,130 @@
...
+  CredentialIssuerMetadataSupportedCredentials,
+  Display,
+  IssuerCredentialSubject,
+  IssuerCredentialSubjectDisplay,
+  SupportedCredentialIssuerMetadataJwtVcJson,
+  SupportedCredentialIssuerMetadataJwtVcJsonLdAndLdpVc,
+  TokenErrorResponse,
+} from '@sphereon/openid4vci-common'
+
+export class CredentialSupportedV1_11Builder {
```
nklomp, line 12: Can we move to suffixes for the Version? It now looks like we are mixing camelcase and snake case. It is clear from the context which type the version applies to.

packages/issuer/lib/functions/CredentialOffer.ts
nklomp: Where is the rename and that actual overload or optional offer object then? Right now you are just creating a state, with no control for the user to define that state. Hance why you really want to be able to also pass in the object, so you have complete control what is in there. Also you are blindly applying supported credential types from the issuer to the offer, which certainly should not happen. The URI needs to be able to only have a subset of what is available in the issuer/AS.

nklomp (changes requested):
See remarks

packages/common/lib/types/OpenID4VCIServerMetadata.ts
sksadjad: changed this to display, because although the text of spec is only describing `name` and `locale`, the example shows the complete `Display` object. So, I've deleted the `NameAndLocaleExtened` and added the catchall param to `NameAndLocale` itself.
also changed the name of `CredentialIssuerMetadataSupportedCredentials` to `CredentialSupported` to be a bit easier to follow

packages/issuer/lib/VcIssuer.ts
sksadjad: added optional parameter for callback function here

packages/issuer/lib/builder/CredentialSupportedV1_11Builder.ts
sksadjad: changed the name of builder from CredentialSupportedV1_11Builder to CredentialSupportedBuilderV1_11

packages/issuer/lib/functions/CredentialOffer.ts
sksadjad: changed this method to meet the criteria that we discussed

packages/common/lib/types/Generic.types.ts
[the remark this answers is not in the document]
sksadjad: changed the `openid4vci_metadata?: OpenID4VCIServerMetadata` to `openid4vci_metadata?: IssuerMetadata`

packages/client/lib/OpenID4VCIClient.ts
```diff
@@ -284,31 +302,60 @@ export class OpenID4VCIClient {
...
+      const supported: IssuerCredentialSubject = {};
+      for (const [key, value] of Object.entries(credentialsSupportedV8)) {
+        if (initiationTypes.includes(key)) {
+          supported[key] = value;
+        }
+      }
+      // todo: fix this later. we're returning CredentialSupportedV1_0_08 as a list of CredentialSupported (for v09 onward)
+      return supported as unknown as CredentialSupported[];
+    }
     if (!credentialsSupported) {
```
nklomp: Suggest to move this to the top. It removes the need for the credentialsSupported && and you are doing the same thing for v8 as well as v8+

packages/issuer/lib/functions/CredentialOfferUtils.ts
```diff
@@ -0,0 +1,38 @@
+import { CredentialOfferPayload, CredentialOfferPayloadV1_0_11, encodeJsonAsURI, IssuerMetadata } from '@sphereon/openid4vci-common'
+import { v4 as uuidv4 } from 'uuid'
+
+export function createCredentialOfferURI(
+  issuerMetadata?: IssuerMetadata,
+  // todo: probably it's wise to create another builder for CredentialOfferPayload that will generate different kinds of CredentialOfferPayload
+  opts?: { state?: string; credentialOffer?: CredentialOfferPayload; preAuthorizedCode?: string; userPinRequired?: boolean }
```
nklomp, line 7: As discussed add an optional credentialOfferURI string value to opts. If present it will be used

Also add an optional scheme string to opts, replacing openid-credential-offer if present

nklomp (changes requested):
Some small remarks

packages/issuer/lib/functions/CredentialOfferUtils.ts
sksadjad: added the `credentialOfferURI` as discussed

packages/client/lib/OpenID4VCIClient.ts
sksadjad: moved at the top of the method and removed the extra checks

nklomp (approved):

PR #40, opened 2023-04-11
feature/VDX-182/jwt_validation

maikel-maas:
- Implemented JWT validation;


packages/client/lib/functions/ProofUtil.ts
```diff
@@ -87,7 +87,7 @@ const createJWT = (jwtProps?: JwtProps, existingJwt?: Jwt): Jwt => {
   };
 
   const jwtHeader: JWTHeaderParameters = {
-    typ: 'JWT',
+    typ: 'openid4vci-proof+jwt',
```
nklomp, line 90: Please do not blindly update these types of values. This is obviously something which should depend on the version of the spec

packages/issuer/lib/VcIssuer.ts
```diff
@@ -57,13 +56,17 @@ export class VcIssuer {
    * @param issueCredentialRequest a credential issuance request
    * @param jwtVerifyCallback OPTIONAL. if provided will use this callback instead what is configured in the VcIssuer
    * @param issuerCallback OPTIONAL. if provided will use this callback instead what is configured in the VcIssuer
+   * @param issuerState the state of the issuer
    */
   public async issueCredentialFromIssueRequest(
     issueCredentialRequest: CredentialRequest,
+    issuerState: string,
```
nklomp, line 63: issuerState is something which is version specific. It was called opState before. Please double check these types of things

packages/issuer/lib/VcIssuer.ts
```diff
@@ -98,10 +101,9 @@ export class VcIssuer {
       }
 
       const { iss, aud, iat, nonce } = payload
-      // Get client id and check if it is pre-authorized
-      if (iss && iss !== '') {
-        throw new Error(ISS_ERROR)
-      }
+      // I couldn't find a way to get the client_id, neither the grant type at this point, so I will not validate iss since it's optional
+      // iss: OPTIONAL (string). The value of this claim MUST be the client_id of the client making the credential request.
+      // This claim MUST be omitted if the Access Token authorizing the issuance call was obtained from a Pre-Authorized Code Flow through anonymous access to the Token Endpoint.
```
nklomp, line 106: You should know whether you are in an authorized or pre-authorized context. In other words whether it should be present or not

packages/issuer-rest/lib/api.ts
```diff
@@ -75,6 +72,8 @@ export class RestAPI {
 
   constructor(opts?: { metadata: IssuerMetadata; stateManager: ICredentialOfferStateManager; userPinRequired: boolean }) {
     dotenv.config()
+    const key = fs.readFileSync(process.env.PRIVATE_KEY || path.join(__dirname, './privkey.pem'), 'utf-8')
```
nklomp, line 75: Why are you adding the notion of keys and certificates to the issuer, which should be completely neutral towards keys?

maikel-maas: One of the requirements of PAR is to use https and that's the reason of having a private key and certificate there. It's not passed to the issuer at any point, only to the web server. It must be called after dotenv.config() otherwise the environment variable is not picked up. Unless you are asking me to remove the private key and certificate from the .env file.

packages/client/lib/functions/ProofUtil.ts
maikel-maas: Fixed!

packages/issuer-rest/lib/api.ts
nklomp: So are we talking about a client certificate here or a server certificate? If the latter it doesn't make sense that the code has any notion of that. If the former, the names are really confusing

packages/issuer/lib/VcIssuer.ts
maikel-maas: I could know from the jwt, if iss is present, it's authorized flow. However, I'm validating the jwt in the credential endpoint and the only thing I have is the jwt, I don't have anything to validate the jwt against. The only thing I can think of is an introspection endpoint in the AS.

nklomp: You could use the state manager to get the credential offer grants. If only one is present that has to be used. If 2 were present then indeed you cannot know whether it should be present or not

packages/issuer-rest/lib/api.ts
nklomp: TLS termination is not the concern of the API/Server. So it simply is out of scope for the code.

If you need it you use a reverse proxy or loadbalancer with certificate. I rather see an option where we allow PAR to not use https in case we are in development mode. So by default it is required and only if this option is enabled we are okay with plain http (which should not be used in production of course)

packages/issuer/lib/VcIssuer.ts
```diff
@@ -100,10 +106,19 @@ export class VcIssuer {
         throw new Error(KID_JWK_X5C_ERROR)
       }
 
-      const { aud, iat, nonce } = payload
-      // I couldn't find a way to get the client_id, neither the grant type at this point, so I will not validate iss since it's optional
+      const { iss, aud, iat, nonce } = payload
       // iss: OPTIONAL (string). The value of this claim MUST be the client_id of the client making the credential request.
       // This claim MUST be omitted if the Access Token authorizing the issuance call was obtained from a Pre-Authorized Code Flow through anonymous access to the Token Endpoint.
+      // TODO We need an introspection endpoint in case the AS and RS are separated
```
nklomp, line 112: Please remove this comment, because we would use access tokens, meaning no overhead of calling the AS from the RS is necessary to begin with. This comment will confuse future developers into the wrong direction

packages/issuer/package.json
```diff
@@ -11,6 +11,7 @@
   "dependencies": {
     "@sphereon/openid4vci-common": "workspace:../common",
     "@sphereon/ssi-types": "^0.9.0",
+    "jose": "^4.10.0",
```
nklomp, line 14: Please move jose to dev deps and copy JwtHeader params from jose to our types in the common module.

I want to stay away from using jose in non-dev imports, as certain jose functions do not work in RN environments.

I know this is the issuer, but the JwtHeader import from jose is in the common lib

packages/client/lib/functions/ProofUtil.ts
```diff
@@ -1,4 +1,13 @@
...
+  JWS_NOT_VALID,
+  Jwt,
+  JWTPayload,
+  ProofOfPossession,
+  ProofOfPossessionCallbacks,
+  ProofType,
+  Typ,
+} from '@sphereon/openid4vci-common';
 import Debug from 'debug';
 import { JWTHeaderParameters } from 'jose';
```
nklomp, line 12: Please remove the  jose import here copy JwtHeader params from jose or from the openid4vci-common as that also already seems to contain Jwt and JWTPayload anyway, it probably also has the concept of Jwt Headers

I want to stay away from using jose in non-dev imports, as certain jose functions do not work in RN environments.

I know this is not part of this PR, but I notice it now and this has the potential to lead to problems down the road.

nklomp (changes requested):
Some small last remarks

packages/issuer-rest/lib/api.ts
maikel-maas: Removed https implementation/support from api.ts

packages/issuer/lib/VcIssuer.ts
maikel-maas: Fixed

packages/issuer/lib/VcIssuer.ts
maikel-maas: Fixed!

packages/issuer/package.json
maikel-maas: Removed!

packages/client/lib/functions/ProofUtil.ts
maikel-maas: Fixed!

packages/common/lib/types/CredentialIssuance.types.ts
```diff
@@ -52,12 +52,34 @@ export type DecodeURIAsJsonOpts = {
 };
 
 export interface JWK {
-  kty?: string;
+  alg?: string;
```
nklomp, line 55: The JWK interface does not need to be extended with all options available. We are not using most of them, so this will only be confusing. Since it is an interface anyway the runtime object could still have additional properties, even if we are not using them.

Suggest to create 2 versions. A BaseJWK only containing the values we use and then a sub interface containing everything if we really want to use it in the future

packages/issuer/lib/VcIssuer.ts
```diff
@@ -55,7 +62,7 @@ export class VcIssuer {
   /**
    * issueCredentialFromIssueRequest
    * @param issueCredentialRequest a credential issuance request
-   * @param preAuthorizedCode the pre-authorized code
+   * @param state the state to be passed to the state manager
```
nklomp, line 65: Update params to be inline with actual method params

packages/issuer/lib/VcIssuer.ts
```diff
@@ -82,7 +85,33 @@ export class VcIssuer {
     throw new Error(TokenErrorResponse.invalid_request)
   }
 
-  private async validateJWT(issueCredentialRequest: CredentialRequest, jwtVerifyCallback?: JWTVerifyCallback, clientId?: string, preAuthorizedCode?: string): Promise<void> {
+  private async retrieveGrantsFromClient(issuerState: string): Promise<{ clientId: string; grants: Grant }> {
+    const credentialOfferState: CredentialOfferState | undefined = await this.retrieveCredentialOffer(issuerState)
+    const clientId = credentialOfferState?.clientId
+    const grants = credentialOfferState?.credentialOffer?.grants
+    if (!clientId) {
+      throw new Error(UNDEFINED_CLIENT_ID)
```
nklomp, line 93: Why are we throwing an error when the optional clientId is not present? It is the first call in the issue method, so a pre-authorized code flow without clientId will fail

packages/issuer/lib/VcIssuer.ts
```diff
@@ -82,7 +85,33 @@ export class VcIssuer {
...
+      throw new Error(GRANTS_MUST_NOT_BE_UNDEFINED)
+    }
+    return { clientId, grants }
+  }
+
+  private async retrieveCredentialOffer(issuerState: string): Promise<CredentialOfferState | undefined> {
+    if (await this._stateManager.hasState(issuerState)) {
+      return await this._stateManager.getState(issuerState)
+    } else {
+      throw new Error(UNKNOWN_CLIENT_ERROR)
```
nklomp, line 105: IMO this should be a method on the stateManager, something like getAssertedState

packages/issuer/package.json
```diff
@@ -11,14 +11,14 @@
...
-    "jose": "^4.10.0",
     "uuid": "^9.0.0"
   },
   "devDependencies": {
     "@sphereon/openid4vci-client": "workspace:../client",
     "@types/jest": "^29.5.0",
     "@types/uuid": "^9.0.1",
-    "nock": "^13.2.9"
+    "nock": "^13.2.9",
+    "jose": "^4.10.0"
```
nklomp, line 21: I would also expect the jose dep to be removed from the common module (or moved to dev deps if it is used in tests)

nklomp (changes requested):
Some last comments

packages/issuer/package.json
maikel-maas: Actually it's from the client module. Fixed

packages/issuer/lib/VcIssuer.ts
maikel-maas: Fixed!

packages/issuer/lib/VcIssuer.ts
maikel-maas: Correct! Removed the error.

packages/issuer/lib/VcIssuer.ts
maikel-maas: Fixed!

packages/common/lib/types/CredentialIssuance.types.ts
maikel-maas: `jose` is picking the properties that we use `Pick<JWK, 'kty' | 'crv' | 'x' | 'y' | 'e' | 'n'>`, but I'm fine with anyone. Fixed!

nklomp (approved):

nklomp:
LGTM

PR #42, opened 2023-04-20
Feature/vdx 187 payloads from issuers

sksadjad:
add:
- metadata, request and responses from multiple different issuers (spruce, walt, uniissuer, mattr, diwala)
- new test for creating credential requests for above issuers
edit:
- types in v8 spec file
fix:
- removed `throw error` if credential request doesn't contain an issuer field
- added v1_0_08 types to exported types of common module


nklomp (approved):
LGTM

PR #43, opened 2023-04-20
VDX-182 Implement V11 Credential Endpoint in OpenID4VCI Issuer minus …

maikel-maas:
…response

   - Fixed JWT expected value failing test;


sksadjad (approved):
LGTM

PR #44, opened 2023-04-20
feature/VDX-189/credential_response

maikel-maas:


packages/common/lib/types/CredentialIssuance.types.ts
```diff
@@ -14,6 +14,9 @@ export interface CredentialRequest {
 export interface CredentialResponse {
   credential: W3CVerifiableCredential;
   format: CredentialFormat | CredentialFormat[];
+  acceptance_token?: string;
```
nklomp, line 17: Okay this is part of the latest spec, but if you read the spec it also made the credential optional as a result of this addition. That is why it is good to have these spec texts about payloads in the interfaces.

So please make the credential optional and also check where that impacts other code

packages/issuer-callback/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@sphereon/openid4vci-issuer-callback",
```
nklomp, line 2: Please rename package and folder to ***callback-example

packages/issuer-callback/lib/IssuerCallback.ts
```diff
@@ -0,0 +1,27 @@
+import * as didKeyDriver from '@digitalcredentials/did-method-key'
+import { Ed25519Signature2020 } from '@digitalcredentials/ed25519-signature-2020'
+import { Ed25519VerificationKey2020 } from '@digitalcredentials/ed25519-verification-key-2020'
+import { securityLoader } from '@digitalcredentials/security-document-loader'
+import vc from '@digitalcredentials/vc'
+import { CredentialRequest } from '@sphereon/openid4vci-common'
+import { ICredential, W3CVerifiableCredential } from '@sphereon/ssi-types'
+
+export const generateDid = async () => {
```
nklomp, line 9: Please add some inline comments that this is just an example and explain what it does

packages/issuer/lib/VcIssuer.ts
```diff
@@ -24,13 +26,16 @@ import {
...
+import { v4 } from 'uuid'
 
 export class VcIssuer {
   _issuerMetadata: IssuerMetadata
   _userPinRequired?: boolean
   _issuerCallback?: CredentialIssuerCallback
   _verifyCallback?: JWTVerifyCallback
   private readonly _stateManager: ICredentialOfferStateManager
+  private readonly _cNonce: string[] = []
+  private readonly _cNonceExpiresIn: number = parseInt(process.env.C_NONCE_EXPIRES_IN as string) * 1000 || 90 * 1000
```
nklomp, line 38: Hmzz, add a todo that we also want some config option in the future next to env vars

packages/issuer/lib/VcIssuer.ts
```diff
@@ -72,9 +77,19 @@ export class VcIssuer {
     const { clientId, grants } = await this.retrieveGrantsFromClient(issuerState)
     await this.validateJWT(issueCredentialRequest, grants, clientId, jwtVerifyCallback)
     if (this.isMetadataSupportCredentialRequestFormat(issueCredentialRequest.format)) {
+      const cNonce = v4()
```
nklomp, line 80: We should probably start adding an optional opts arg with an optional nonce value. Use that in case it is present, otherwise generate it

packages/issuer/lib/VcIssuer.ts
```diff
@@ -24,13 +26,16 @@ import {
...
 import { ICredential, W3CVerifiableCredential } from '@sphereon/ssi-types'
+import { v4 } from 'uuid'
 
 export class VcIssuer {
   _issuerMetadata: IssuerMetadata
   _userPinRequired?: boolean
   _issuerCallback?: CredentialIssuerCallback
   _verifyCallback?: JWTVerifyCallback
   private readonly _stateManager: ICredentialOfferStateManager
+  private readonly _cNonce: string[] = []
```
nklomp, line 37: Plural, also we need to create an interface for a nonce manager or something. Since this needs to be extended in the future with a persistent store, similar like for the credential offer state manager.

It is not the concern of the low level lib to provide all kinds of implementations, but we should be able to start doing proper persistence for instance in our SDK

packages/issuer/lib/__tests__/VcIssuer.spec.ts
```diff
@@ -12,7 +12,8 @@ import {
   ProofOfPossession,
   Typ,
 } from '@sphereon/openid4vci-common'
-import { IProofPurpose, IProofType } from '@sphereon/ssi-types'
+import { generateDid, getIssuerCallback } from '@sphereon/openid4vci-issuer-callback/dist/IssuerCallback'
```
nklomp, line 15: Import is not correct. Also not sure why the issuer lib would depend on the issuer-callback lib instead of the other way around

packages/issuer/package.json
```diff
@@ -15,7 +15,9 @@
   },
   "devDependencies": {
     "@sphereon/openid4vci-client": "workspace:../client",
+    "@sphereon/openid4vci-issuer-callback": "workspace:../issuer-callback",
```
nklomp, line 18: I would expect the deps to be the other way around

nklomp (changes requested):
See remarks

packages/common/lib/types/CredentialIssuance.types.ts
maikel-maas: Fixed!

packages/issuer-callback/package.json
maikel-maas: Fixed!

packages/issuer-callback/lib/IssuerCallback.ts
maikel-maas: Fixed!

packages/issuer/lib/VcIssuer.ts
maikel-maas: Fixed!

packages/issuer/lib/VcIssuer.ts
maikel-maas: Fixed!

packages/issuer/package.json
maikel-maas: Yes, I've made it issuer centric. Fixed

packages/callback-example/lib/__tests__/issuerCallback.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+  Typ,
+} from '@sphereon/openid4vci-common'
+import { CredentialSupportedBuilderV1_11, VcIssuer, VcIssuerBuilder } from '@sphereon/openid4vci-issuer'
+import { MemoryCredentialOfferStateManager } from '@sphereon/openid4vci-issuer/dist/state-manager/MemoryCredentialOfferStateManager'
+import { ICredential, IProofPurpose, IProofType, W3CVerifiableCredential } from '@sphereon/ssi-types'
+import * as jose from 'jose'
+
+import { generateDid, getIssuerCallback, verifyCredential } from '../IssuerCallback'
+
+const INITIATION_TEST_URI =
```
nklomp, line 21: Why is the issuer using a v8/v9 URI, whilst the Issuer itself is v11+

packages/issuer/lib/VcIssuer.ts
```diff
@@ -24,31 +27,37 @@ import {
...
-      stateManager: ICredentialOfferStateManager
+      stateManager: IStateManager<CredentialOfferState>
+      cNonceStateManager: IStateManager<CNonceState>
       callback?: CredentialIssuerCallback
       verifyCallback?: JWTVerifyCallback
     }
   ) {
     this._issuerMetadata = issuerMetadata
     this._stateManager = args.stateManager
+    this._cNonceStateManager = args.cNonceStateManager
```
nklomp, line 54: Just call that nonceManager. It becomes very confusing is you call one thing a state manager and then another a cnoncestate manager

packages/issuer/tsconfig.json
```diff
@@ -8,6 +8,9 @@
   "references": [
     {
       "path": "../common"
+    },
+    {
+      "path": "../callback-example"
```
nklomp, line 13: Why is this reference still present?

packages/common/lib/types/CredentialIssuance.types.ts
```diff
@@ -12,8 +12,11 @@ export interface CredentialRequest {
 }
 
 export interface CredentialResponse {
-  credential: W3CVerifiableCredential;
-  format: CredentialFormat | CredentialFormat[];
+  credential?: W3CVerifiableCredential; // OPTIONAL. Contains issued Credential. MUST be present when acceptance_token is not returned. MAY be a JSON string or a JSON object, depending on the Credential format. See Appendix E for the Credential format specific encoding requirements
```
nklomp, line 15: Since you now made credential optional, I would expect an expression somewhere checking the value to be present before V11, as it was required back than. Don't just change optionality without actually doing something with it if it changed between versions

nklomp (changes requested):
See questions/remarks

packages/common/lib/types/CredentialIssuance.types.ts
maikel-maas: Acceptance_token is part of the deferred response, so for now I will only validate that the credential is actually there (mandatory since deferred response is not implemented).

packages/callback-example/lib/__tests__/issuerCallback.spec.ts
maikel-maas: Fixed!

packages/issuer/lib/VcIssuer.ts
maikel-maas: Fixed!

packages/issuer/tsconfig.json
maikel-maas: Fixed!

packages/callback-example/lib/__tests__/issuerCallback.spec.ts
nklomp: That was not the point. The point was openid-initiate-issuance vs openid-credential-offer, so v8/9 vs v11.
But since the client does not fully support v11 for now, I guess it is okay for now

nklomp (approved):
LGTM

PR #45, opened 2023-04-26
feature/VDX-193/token_endpoint

maikel-maas:


packages/common/lib/types/StateManager.types.ts
```diff
@@ -4,6 +4,11 @@ export interface CredentialOfferState {
   clientId?: string;
   credentialOffer: CredentialOfferPayloadV1_0_11;
   createdOn: number;
+  'pre-authorized_code': string;
```
nklomp, line 7: Why is this stored separately? This is part of the Grant in the credentialOffer

packages/common/lib/types/StateManager.types.ts
```diff
@@ -4,6 +4,11 @@ export interface CredentialOfferState {
...
+  preAuthorizedCodeCreatedOn: number;
```
nklomp, line 8: Why is this property needed if there already is a createdOn?

packages/common/lib/types/StateManager.types.ts
```diff
@@ -4,6 +4,11 @@ export interface CredentialOfferState {
...
+  preAuthorizedCodeExpiresIn: number;
```
nklomp, line 9: I rather have a configurable expiration duration, which we use together with the createdOn to determine whether a token is expired. Means we do not need to store this property at all.

packages/common/lib/types/StateManager.types.ts
```diff
@@ -4,6 +4,11 @@ export interface CredentialOfferState {
...
+  userPin: number;
```
nklomp, line 10: Why are the userPin properties stored if they are already part of the credentialOffer?

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
```diff
@@ -0,0 +1,134 @@
...
+let tokenPath: string
+let interval: number
+let cNonceExpiresIn: number
+let tokenExpiresIn: number
+let stateManager: IStateManager<CredentialOfferState>
+let nonceStateManager: IStateManager<CNonceState>
+let jwtSignerCallback: JWTSignerCallback
+
+export const tokenRequestEndpoint = (opts?: ITokenEndpointOpts): Router => {
+  tokenPath = (opts?.tokenPath ? opts.tokenPath : (process.env.TOKEN_PATH as string)) ?? '/token'
```
nklomp, line 36: Interesting you are mixing a ternary with nullish coalescing. Why not use nullish coalescence everywhere then?
````typescript
opts?.tokenPath ?? process.env.TOKEN_PATH as string ?? '/token'
````

File: packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts [hunk 6]
```diff
@@ -0,0 +1,134 @@
...
+  interval = (opts?.interval ? opts.interval : getNumberOrUndefined(process.env.INTERVAL)) ?? 300000
```
nklomp, line 37: Same remark as above for this mixture of ternary and nullish coalescence (same for the next 2 lines)

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
```diff
@@ -0,0 +1,134 @@
...
+  cNonceExpiresIn = (opts?.cNonceExpiresIn ? opts.cNonceExpiresIn : getNumberOrUndefined(process.env.C_NONCE_EXPIRES_IN)) ?? 300000
+  tokenExpiresIn = (opts?.tokenExpiresIn ? opts.tokenExpiresIn : getNumberOrUndefined(process.env.TOKEN_EXPIRES_IN)) ?? 300000
+  if (opts?.jwtSignerCallback) {
```
nklomp, line 40: Suggest to rename this to accessTokenSignerCallback. A JWT is a generic concept. What if we need other token types in the future?

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
```diff
@@ -0,0 +1,134 @@
...
+  getNumberOrUndefined,
+  IStateManager,
+  Jwt,
+  JWTSignerCallback,
+  Typ,
+} from '@sphereon/openid4vci-common'
+import express, { NextFunction, Request, Response, Router } from 'express'
+import { v4 } from 'uuid'
+
+const router = express.Router()
```
nklomp, line 15: What happened to the requirement that we should be able to pass in an existing express object?

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
```diff
@@ -0,0 +1,134 @@
...
+  }
+  router.post(tokenPath, handleHTTPStatus400, handleTokenRequest)
+  return router
+}
+
+const generateAccessToken = async (): Promise<string> => {
+  const issuanceTime = new Date()
+  const jwt: Jwt = {
+    header: { typ: Typ.JWT, alg: Alg.ES256 },
+    payload: { iat: issuanceTime.getTime(), exp: +new Date(issuanceTime.getTime() + tokenExpiresIn) },
```
nklomp, line 63: I would expect a relation to the actual credential offer somehow in the access token, so we can get back to the state. This is a JWT signed by us, so it makes sense that we will trust it coming back in as long as the signature is valid and from us

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
```diff
@@ -0,0 +1,134 @@
...
+}
+
+const handleTokenRequest = async (request: Request, response: Response) => {
+  response.set({
+    'Cache-Control': 'no-store',
+    Pragma: 'no-cache',
+  })
+
+  const cNonce = v4()
+  await nonceStateManager.setState(cNonce, { cNonce, createdOn: +new Date() })
```
nklomp, line 75: Please use seperate values for the nonce (OpenID) and state (OAuth2) values. I am fine with one being the has of the other, or just a seperate uuid.

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
```diff
@@ -0,0 +1,134 @@
...
+  setTimeout(() => {
+    nonceStateManager.deleteState(cNonce)
```
nklomp, line 77: This should only be done if the memory nonce manager is used and IMO is a concern of that manager itself, not of the calling code. For instance a persistent store would probably run something in the background to cleanup. This approach would result in the issuer keeping track of an awful lot of timeouts

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
```diff
@@ -0,0 +1,134 @@
...
+    c_nonce: cNonce,
+    c_nonce_expires_in: cNonceExpiresIn,
+    authorization_pending: false,
+    interval,
+  }
+  return response.status(200).json(responseBody)
+}
+
+export const handleHTTPStatus400 = async (request: Request, response: Response, next: NextFunction) => {
+  // TODO invalid_client: the client tried to send a Token Request with a Pre-Authorized Code without Client ID but the Authorization Server does not support anonymous access
```
nklomp, line 94: What does this remark mean? There is no AS yet

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
```diff
@@ -0,0 +1,134 @@
...
+  const assertedState = (await stateManager.getAssertedState(request.body.state)) as CredentialOfferState
+  await stateManager.setState(request.body.state, assertedState)
```
nklomp, line 96: Wait what? Why are we updating the state which we just got with a value coming in from the client, which we do not trust to begin with. Doesn't make any sense to me

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
```diff
@@ -0,0 +1,134 @@
...
+    invalid_grant:
+    the Authorization Server expects a PIN in the pre-authorized flow but the client provides the wrong PIN
+    the End-User provides the wrong Pre-Authorized Code or the Pre-Authorized Code has expired
+     */
+    if ((request.body.user_pin && !/[0-9{,8}]/.test(request.body.user_pin)) || assertedState.userPin != getNumberOrUndefined(request.body.user_pin)) {
+      return response.status(400).json({ error: 'invalid_grant', error_message: 'PIN must consist of maximum 8 numeric characters' })
+    }
+    const now = +new Date()
+    const expirationTime = assertedState.preAuthorizedCodeCreatedOn + assertedState.preAuthorizedCodeExpiresIn
+    if (
```
nklomp, line 125: See remarks about storing these in the state itself (most of them are already there).

I don't like this expression and the error message. These are completely seperate concerns, so why do this in one expression.

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
```diff
@@ -0,0 +1,134 @@
...
+    interval,
+  }
+  return response.status(200).json(responseBody)
+}
+
+export const handleHTTPStatus400 = async (request: Request, response: Response, next: NextFunction) => {
+  // TODO invalid_client: the client tried to send a Token Request with a Pre-Authorized Code without Client ID but the Authorization Server does not support anonymous access
+  const assertedState = (await stateManager.getAssertedState(request.body.state)) as CredentialOfferState
+  await stateManager.setState(request.body.state, assertedState)
+  if (request.body.grant_type === 'urn:ietf:params:oauth:grant-type:pre-authorized_code') {
```
nklomp, line 97: This should also be checked against the actual grant types in the credential offer from the state manager

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
```diff
@@ -0,0 +1,134 @@
...
+  tokenPath?: string
+  interval?: number
+  cNonceExpiresIn?: number
+  tokenExpiresIn?: number
+  stateManager?: IStateManager<CredentialOfferState>
+  nonceStateManager?: IStateManager<CNonceState>
+  jwtSignerCallback?: JWTSignerCallback
+}
+
+let tokenPath: string
```
nklomp, line 27: Why is this using global state variables?

packages/issuer-token-endpoint/lib/IssuerTokenServer.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+    this._app.use(
+      this._baseUrl.pathname,
+      tokenRequestEndpoint({ tokenPath, tokenExpiresIn, interval, cNonceExpiresIn, stateManager, nonceStateManager, jwtSignerCallback })
+    )
+  }
+}
+
+const signerCallback = async (jwt: Jwt, kid?: string): Promise<string> => {
+  const privateKey = (await jose.generateKeyPair(Alg.ES256)).privateKey as KeyObject
```
nklomp, line 57: Why is the implementation using a predefined and random signer callback? That is what needs to be provided. Something completely random will never work of course

packages/issuer-token-endpoint/lib/IssuerTokenServer.ts
```diff
@@ -0,0 +1,82 @@
...
+  return new jose.SignJWT({ ...jwt.payload }).setProtectedHeader({ ...jwt.header }).sign(privateKey)
+}
+
+const state = new MemoryCredentialOfferStateManager()
+state.setState('test_state', {
```
nklomp, line 62: What is this code doing here?

nklomp (changes requested):
See remarks.

IMO too many remarks, that should have been discussed first.

packages/issuer-token-endpoint/lib/IssuerTokenServer.ts
maikel-maas: Removed

packages/issuer-token-endpoint/lib/IssuerTokenServer.ts
maikel-maas: Removed

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
maikel-maas: Removed!

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
maikel-maas: Fixed!

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
maikel-maas: Fixed!

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
maikel-maas: I'm rocker but never got to know Elvis' work. :laughing: Now I see it's possible to chain them.

packages/common/lib/types/StateManager.types.ts
maikel-maas: We have `user_pin_required` inside `GrantUrnIetf` but no `user_pin`. Should I add it to `GrantUrnIetf`?

packages/common/lib/types/StateManager.types.ts
maikel-maas: This is actually duration, I'm just using the suffix expires in for the sake of consistency. BTW, I think we can read it as 'pre-authorized_code expires in 1 minute' or something. I'm not sure if it is semantically wrong.

packages/common/lib/types/StateManager.types.ts
maikel-maas: Fixed!

packages/common/lib/types/StateManager.types.ts
maikel-maas: Fixed!

nklomp:
Please address all comments

nklomp (changes requested):
Please address all comments

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
maikel-maas: Just garbage left behind

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
maikel-maas: Removed!

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
maikel-maas: Fixed!

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
maikel-maas: Fixed!

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
maikel-maas: I'm binding the state to the iss property. I think it's what makes more sense.

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
maikel-maas: This is just a router to add flexibility on how the baseUrl is handled. The IssuerTokenEndpoint can be plugged on an existing express server **or** an express instance can be passed to IssuerTokenServer and they will work the same way.

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
maikel-maas: Fixed!

packages/common/lib/types/StateManager.types.ts
maikel-maas: There is a validation rule:

invalid_grant:
      the Authorization Server expects a PIN in the pre-authorized flow but the client provides the wrong PIN
      the End-User provides the wrong Pre-Authorized Code or the **Pre-Authorized Code has expired**

Since the token expiration time not present in the [request](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-token-request), How should I check whether it is expired?

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
```diff
@@ -0,0 +1,172 @@
...
+  return router
+}
+
+const generateAccessToken = async (
+  opts: Required<Pick<ITokenEndpointOpts, 'accessTokenSignerCallback' | 'tokenExpiresIn'> & { state: string }>
+): Promise<string> => {
+  const issuanceTime = new Date()
+  const jwt: Jwt = {
+    header: { typ: Typ.JWT, alg: Alg.ES256 },
+    payload: { iat: issuanceTime.getTime(), exp: opts.tokenExpiresIn, iss: opts.state },
```
nklomp, line 76: What is the state doing in the iss field (that is the field for an identity of the issuer?

nklomp (changes requested):
See remarks.

Don't abuse the iss field. That is the identity of an issuer. It for sure should not contain a state value.

packages/issuer-token-endpoint/lib/IssuerTokenEndpoint.ts
maikel-maas: Fixed!

packages/common/lib/types/StateManager.types.ts
maikel-maas: Fixed!

nklomp (approved):

PR #50, opened 2023-05-24
fix: authorization parameter encoding not as JSON

blu3beri:
Small fix for encoding the Authorization Request.

The spec shows the example as:

```
GET /authorize?
  response_type=code
  &client_id=s6BhdRkqt3
  &code_challenge=E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM
  &code_challenge_method=S256
  &authorization_details=%5B%7B%22type%22:%22openid_credential
  %22,%22format%22:%22jwt_vc_json%22,%22types%22:%5B%22Verifia
  bleCredential%22,%22UniversityDegreeCredential%22%5D%7D%5D
  &redirect_uri=https%3A%2F%2Fclient.example.org%2Fcb
Host: https://server.example.com
```

which encodes each key-value pair inside the JSON object as separate query parameters. Right now, it is encoded as the entire JSON body as a key and an empty string as value. 

The fix I used here is not the cleanest but it works as a simple override with the encoding algorithm used.

### DIFF
--- a/packages/client/lib/OpenID4VCIClient.ts
+++ b/packages/client/lib/OpenID4VCIClient.ts
@@ -144,6 +144,7 @@ export class OpenID4VCIClient {
       baseUrl: this._endpointMetadata.authorization_endpoint,
       uriTypeProperties: ['redirect_uri', 'scope', 'authorization_details'],
       version: this.version(),
+      encodeAsSingleJson: false,
     });
   }
 
--- a/packages/client/lib/functions/index.ts
+++ b/packages/client/lib/functions/index.ts
@@ -1,3 +1,3 @@
 export * from '@sphereon/oid4vci-common/dist/functions/Encoding';
-export * from '../../../common/lib/functions/HttpUtils';
+export * from '@sphereon/oid4vci-common/dist/functions/HttpUtils';
 export * from './ProofUtil';
--- a/packages/common/lib/functions/Encoding.ts
+++ b/packages/common/lib/functions/Encoding.ts
@@ -32,7 +32,7 @@ export function convertJsonToURI(
   }
 
   let components: string;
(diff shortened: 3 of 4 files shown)


nklomp:
Hi @blu3beri thanks for the PR. I will look into this later today. If I am not mistaken this is actually what it should be since v10/v11. Or let met rephrase. They moved it from seperate params to a json value, which should be encoded in it's entirety.

nklomp:
It changed between v8 and v9 apparently. So we need to update the expression a bit to only apply it for < v9.

See 
v8: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-08.html#name-issuance-initiation-request
v11: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-4.1.2


I would be okay to accept the PR, but at the same time IMO it makes more sense to do the right thing automatically based on the actual deducted version of the credential offer

blu3beri:
> I would be okay to accept the PR, but at the same time IMO it makes more sense to do the right thing automatically based on the actual deducted version of the credential offer

I think we should do it do correct way based on the version. Would be happy to pick this up myself or do you have an idea of the solution?

nklomp:
In this branch the version is already returned when getting the credential offer.

Then when that version would be passed in to this method it should do the correct thing. The only change should be that instead of checking against < v11 that should be changed to < v9.
Also the current method expects v11 by default, so if you do not pass in the version it will also do the encoding of the complete JSON object. So the expression needs to be changed from v11 to v9 and then it is a matter of making sure the version gets passed in

nklomp:
I will probably not be able to look at it before somewhere later tomorrow though

nklomp:
This should be fixed in latest version without the param. If not, please reopen

PR #52, opened 2023-06-01
fix: rename jwt_vc_json_ld to jwt_vc_json-ld and fixed test

blu3beri:
According to https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-e.1 the format identifier for a VC signed as a JWT, using JSON-ld is `jwt_vc_json-ld` and not `jwt_vc_json_ld`.

Also, this fixes a small mistake, I assume, in the test where an extra 'Q' was added to the check.


nklomp:
Hi @blu3beri thanks, yeah makes sense. Will have a look at it later.

The Q wasn't an error as all the tests pass. Because you changed the value it impacts the base64url encoding for the JWT, meaning the Q isn't valid anymore.

PR #53, opened 2023-06-01
fix: rename jwt_vc_json_ld to jwt_vc_json-ld

blu3beri:
According to https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-e.1 the format identifier for a VC signed as a JWT, using JSON-ld is `jwt_vc_json-ld`.

(a better fix might be to keep the format identifiers in an enum and reuse them form there.)

### DIFF
--- a/packages/client/lib/CredentialRequestClient.ts
+++ b/packages/client/lib/CredentialRequestClient.ts
@@ -58,7 +58,7 @@ export class CredentialRequestClient {
       let format: string = uniformRequest.format;
       if (format === 'jwt_vc_json') {
         format = 'jwt_vc';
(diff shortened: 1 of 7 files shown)


nklomp:
Hi @blu3beri thanks, yeah makes sense. Will have a look at it later.

The Q wasn't an error as all the tests pass. Because you changed the value it impacts the base64url encoding for the JWT, meaning the Q isn't valid anymore.

Regarding to move to an enum. We typically dislike enums for things that end up in payloads. It already is a type, meaning it will only accept the values listed for the type, which removed potential for errors. It currently might also have a union with the 'string' type, which we will remove

nklomp (approved):

PR #54, opened 2023-06-05
feat: inline credential offers and credential supported mapping improvements

TimoGlastra:
This PR adds support for inline credential offers in the `credentials` property of an v11 credential offer. An inline offer does not map to any supported credential in the issuer metadata.

With the new methods etc. how you can use it is as follows:
- call `getOfferedCredentialsWithMetadata` on the client. This returns a list of `CredentialSupported` or `CredentialOfferFormat`.
- call `acquireCredentials...` and you provide the `types` and `format` from either the `CredentialsSupported` or the `CredentialOfferFormat`.

This prevents a lot of version checking and having to access a lot of internal data structures on our side.

Some notes on the implementation:
- We could try to fit the `CredentialOfferFormat` into something like a `CredentialSupported` structure (a common structure between both) to make it easier for the user of the library to work with inline offers, but I think this would only add complexity to the interface of the library, and there's no guarantee for the objects to have overlap.
- I'm quite confused by inline credential offers, as I don't see how you can infer what crypto suites, did methods, etc.. are supported by the issuer for an inline credential offer.
- I'm quite confused by the way a credential request works as you request based on the `types` and `format` (which are both present in the `CredentialSupported` and `CredentiaOfferFormat`). What I don't get is why the credential request (this is about the spec) doesn't just accept an `id` of the credential as is used in the credentials supported. This makes it quite ambiguous which credential you're requesting IMO.

I've also made some improvements to the mapping of the v8 structure to v11 and included an `id` for all the objects based on the key that was used in v8.

So a v8 structure of this:

```ts
OpenBadgeCredential: {
        formats: {
          ldp_vc: {
            name: 'JFF x vc-edu PlugFest 2',
            description: "MATTR's submission for JFF Plugfest 2",
            types: ['OpenBadgeCredential'],
            binding_methods_supported: ['did'],
            cryptographic_suites_supported: ['Ed25519Signature2018'],
          },
        jwt_vc: {
            name: 'JFF x vc-edu PlugFest 2',
            description: "MATTR's submission for JFF Plugfest 2",
            types: ['OpenBadgeCredential'],
            binding_methods_supported: ['did'],
            cryptographic_suites_supported: ['EdDSA'],
          },
        },
      },
```

is now transformed to (approximation):

```ts
[
 {
           id: 'OpenBadgeCredential-ldp_vc',
            name: 'JFF x vc-edu PlugFest 2',
            description: "MATTR's submission for JFF Plugfest 2",
            types: ['OpenBadgeCredential'],
            binding_methods_supported: ['did'],
            cryptographic_suites_supported: ['Ed25519Signature2018']
},
{
           id: 'OpenBadgeCredential-jwt_vc',
            name: 'JFF x vc-edu PlugFest 2',
            description: "MATTR's submission for JFF Plugfest 2",
            types: ['OpenBadgeCredential'],
            binding_methods_supported: ['did'],
            cryptographic_suites_supported: ['EdDSA']
}
]
```

This makes it possible to find the supported credential based on the id used in the credential offer. To help with the id having a suffix, there's a new `getCredentialsSupportedById`, which takes this id conflict into account.

Furthermore, there seemed to be a mix up of the credential types (between the `types` property  and the `id` from credentials offered/supported). This is my attempt to fix it.

I haven't added tests as I first want some feedback on the approach.

### DIFF
--- a/packages/client/lib/OpenID4VCIClient.ts
+++ b/packages/client/lib/OpenID4VCIClient.ts
@@ -10,6 +10,8 @@ import {
   CredentialResponse,
   CredentialSupported,
   EndpointMetadata,
+  OfferedCredentialsWithMetadata,
+  OfferedCredentialType,
   OpenId4VCIVersion,
   OpenIDResponse,
   ProofOfPossessionCallbacks,
@@ -405,8 +407,8 @@ export class OpenID4VCIClient {
    * from this method could contain multiple entries for a single credential id, but with different formats. This is detectable as the
    * id will be the `<credentialId>-<format>`.
    */
-  getOfferedCredentialsWithMetadata(): Array<CredentialSupported | CredentialOfferFormat> {
-    const offeredCredentials: Array<CredentialSupported | CredentialOfferFormat> = [];
+  getOfferedCredentialsWithMetadata(): Array<OfferedCredentialsWithMetadata> {
+    const offeredCredentials: Array<OfferedCredentialsWithMetadata> = [];
 
     for (const offeredCredential of this.getOfferedCredentials()) {
       // If the offeredCredential is a string, it references a supported credential in the issuer metadata
@@ -418,7 +420,11 @@ export class OpenID4VCIClient {
           throw new Error(`Offered credential '${offeredCredential}' is not present in the credentials_supported of the issuer metadata`);
         }
 
-        offeredCredentials.push(...credentialsSupported);
+        offeredCredentials.push(
+          ...credentialsSupported.map((credentialSupported) => {
+            return { credentialSupported, type: OfferedCredentialType.CredentialSupported } as const;
+          })
+        );
       }
       // Otherwise it's an inline credential offer that does not reference a supported credential in the issuer metadata
       else {
@@ -438,7 +444,7 @@ export class OpenID4VCIClient {
         //      types: offeredCredential.credential_definition.types,
         //    } satisfies CredentialSupported;
         //  }
-        offeredCredentials.push(offeredCredential);
+        offeredCredentials.push({ inlineCredentialOffer: offeredCredential, type: OfferedCredentialType.InlineCredentialOffer } as const);
       }
     }
 
--- a/packages/client/lib/OpenID4VCIClient.ts
+++ b/packages/client/lib/OpenID4VCIClient.ts
@@ -4,10 +4,12 @@ import {
   AuthorizationRequestV1_0_09,
   AuthzFlowType,
   CodeChallengeMethod,
+  CredentialOfferFormat,
   CredentialOfferPayloadV1_0_08,
   CredentialOfferRequestWithBaseUrl,
   CredentialResponse,
   CredentialSupported,
+  CredentialSupportedTypeV1_0_08,
   EndpointMetadata,
   OID4VCICredentialFormat,
   OpenId4VCIVersion,
@@ -17,7 +19,6 @@ import {
   ResponseType,
 } from '@sphereon/oid4vci-common';
 import { getSupportedCredentials } from '@sphereon/oid4vci-common/dist/functions/IssuerMetadataUtils';
-import { CredentialSupportedTypeV1_0_08 } from '@sphereon/oid4vci-common/dist/types/v1_0_08.types';
 import { CredentialFormat } from '@sphereon/ssi-types';
 import Debug from 'debug';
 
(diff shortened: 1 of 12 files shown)


packages/client/lib/OpenID4VCIClient.ts
```diff
@@ -359,12 +360,42 @@ export class OpenID4VCIClient {
...
+   *
+   * NOTE: for v1_0-08, a single credential id in the issuer metadata could have multiple formats. When retrieving the
+   * supported credentials (using {@link getCredentialsSupported}), for v1_0-08, the format is appended to the id if
+   * there are multiple formats supported for that credential id. E.g. if the issuer metadata for v1_0-08 contains an
+   * entry with key `OpenBadgeCredential` and the supported formats are `jwt_vc-jsonld` and `ldp_vc`, then the id
+   * in the credentials supported will be `OpenBadgeCredential-jwt_vc-jsonld` and `OpenBadgeCredential-ldp_vc`, even though
+   * the offered credential is simply `OpenBadgeCredential`. This method takes this into account.
+   */
+  getCredentialsSupportedById(credentialSupportedId: string, format?: CredentialFormat): CredentialSupported[] {
+    return this.getCredentialMetadata(credentialSupportedId).filter((credentialSupported) => credentialSupported.format === format);
```
nklomp, line 375: What is the intent if the format is not supplied? I would expect it to return anything regardless of format

nklomp (changes requested):
Some small initial remarks. IMO we need to make sure that we throw an error in case we have a offer containing a string for the credential, but the actual credential supported not resolving

TimoGlastra:
>  I need to look at the spec though as probably some array values in the Credentials Support for crypto/key and method bindings would prevent it from uniquely identifying a unique crypto binding, format and credential type.

Yeah I'm really not sure how they envision the inline offers to detect which dids methods etc.. are available.

> Regarding you confusion about inline credential offers: IMO these should be similar to CredentialsSupported, because as you mentioned you are missing info to determine what DID methods/key types to use etc.

So that should be changed in the spec then right? There's not really something we can do right about that it seems? I can map them to the `CredentialsSupported` structure, which means the api is consistent between the two. But the only values we'll be able to populate are `format`, `types` and `@context`. 

I think for now it may be good to return them as a separate object, as the handling of inline offers is really ambiguous if you ask me.

TimoGlastra:
I'm having a hard time wrapping up the implementation as I think there's a mixing of the types in this library. Basically there's two types:
- `types` arrays which reflect types from a vc types array (e.g. `VerifiableCredential`). 
- the type of the openid4vci credential being offered, also known as the `id` property in in the `credential_supported`.


When a credential is being requested, it looks whether the credentialTypes is availalbe in the `types` of a credential supported. 

The library mentions that the `credentialType` "Needs to match a type from the Initiate Issance Request!". 

It seems this has changed between draft 8 and draft 11. I'll try to highlight the differences in flows here:

**draft 8**
- Credential offer contains `credential_type` string that points to a credential in the credential supported metadata
- Credential request contains a `type` property that must be the same as the `credential_type` in the offer and points to a credential in the credential supported metadata

Now in draft 11 this completely changed, and there's now:

- Credential offer contains `credentials` array that contains a list of strings (that point to a credential in the credentials supported metadata) or an inline credential offer containing `format` and format specific properties.
- Credential request is dependant based on the `format` and contains a `types` array for w3c credentials, but not for e.g. the iso format. You don't provide the id from the credential offer / credential issuer metadata anymore, but rather you inline some property related to that format. This also allows you to use the new inline credential offer, which contains those properties needed for the request directly.

What this had lead to is overloading of the `type` and `credentialType` properties, where there's a mix between the `credential_type` from draft 8, and the `types` now being used in draft 11.

I think this requires a bit more refactoring of the code, that when requesting a credential you either provide a `supportedCredentialId` (We can name it whatever, but I hope this will clear up some ambiguity) OR we require the request properties to be passed directly based on the `CredentialOfferFormat` (inline credential offer).

TimoGlastra:
Okay. So I think I've got it mostly handled right now and most tests are passing again.

I've separated all the different usages of `types` vs `type`, and it should now be according to the spec. It's a bit weird as you don't include the `id` from the `credentials_supported` in draft 9+ (but there's an ongoing discussion: https://bitbucket.org/openid/connect/issues/1922/oid4vci-unique-id-for-each-element-in). 

To accomodate for all different ways of offering and requesting a credential, I've added the following ways:
- request using `credentialSupported`. This must either be an object with `id` and `format`. This is mostly for a bit easier backwards compatiblity with v8, so you don't have to pass the whole `CredentialSupported` object. But we could remove it and just require you to pass the whole `CredentialSupported` (but if you just want to include the `type` and `format`, this may be nice to have). Or it can be a full `CredentialSupported` and it will extract the needed properties. For draft 9+ you MUST pass the whole `CredentialSupported` as it is needed to extract properties from it (such as the `types` array , or the `@context` property).
- request using `inlineCredentialOffer`. This must be an object as passed in the  `credentials` array from the credential offer. Based on the object, a credential request for that specific format will be constructed. This method requires the draft to be 9+ as draft 8 did not have inline offers yet.
- request using `requestInput`. This allows you to construct the parameters for the request directly. This is also only availalbe for draft 9+ as from then on the request format changed significantly and it didn't contain a pointer anymore to `credentials_supported` that draft 8 did.

I think this way we can support all flows and make sure we keep flexibility.

One thing I ran into when fixing the tests is that I added a few checks based on versions, and it seems some tests have e.g. draft 8 initiation uris, but expect the output request to be draft 9. It seems to be based on a lot of JFF issues, so I'm wondering whether I'm missing something here? Is there an in-between version that was used for these issuers that is not draft 8 and not draft 9? Or is there a mix-up in the tests here?


The new `getOfferedCredentialsWithMetadata` allows you to handle both inline offers and offers pointing to credentials_supported from the same method:
 
```ts
    // Loop through all offered credentials
    for (const offeredCredential of client.getOfferedCredentialsWithMetadata()) {
      // Type allows to easily check the type of the offered credentials (CredentialSupported or InlineCredentialOffer)
      const { type, ...offeredCredentialData } = offeredCredential;

      const credentialResponse = await client.acquireCredentials({
        // You can easily spread the offeredCredentialData into the acquireCredentials call
        // This hides the complexity of the different types of offered credentials (inline vs credential supported),
        // but it still gives you full access to the data of the offered credential if you want
        ...offeredCredentialData,
        proofCallbacks: {
          signCallback: proofOfPossessionCallbackFunction,
        },
      });
    }
```

nklomp:
Overall looks like a good improvement. I need to look into the specifics a bit more and noticed some small issues with CI failing. Did you notice these?

TimoGlastra:
> noticed some small issues with CI failing. Did you notice these?

Yeah so these issues have to do with some tests that I think are incorrect (from previous message):

One thing I ran into when fixing the tests is that I added a few checks based on versions, and it seems some tests have e.g. draft 8 initiation uris, but expect the output request to be draft 9. It seems to be based on a lot of JFF issues, so I'm wondering whether I'm missing something here? Is there an in-between version that was used for these issuers that is not draft 8 and not draft 9? Or is there a mix-up in the tests here?

nklomp:
It could be a mixup on our end and that JFF really was about v8 to begin with. At least that is what I am expecting ATM

genaris:
This is probably an inappropriate question here and you'll want to throw something over my head but I would like how far we are now from supporting these inline credentials. I've been playing with current [Launchpad from MATTR](https://launchpad.mattrlabs.com/) trying to get a VC for the upcoming JFF PlugFest 3 and the QR code is something like this:

```
openid-credential-offer://?credential_offer={"credential_issuer":"https://launchpad.vii.electron.mattrlabs.io","credentials":[{"format":"ldp_vc","types":["OpenBadgeCredential"]}],"grants":false}
```

Is it something else needed apart from this PR to get this flow working? I'd be glad to help if needed!

nklomp:
Hi @genaris Thanks for bringing it up. We for sure will be looking into this soon.

Although Mattr not working doesn't seem to have a relation with this PR at first sight. This PR is mainly about allowing inline credential definitions in the QR or in the credential offer URI. That is not what Mattr is using. Their QR looks bigger, simply because they do not use a URI for the credential offer, meaning they have to cram everything regarding the offer into the QR.

Our wallet is also failing on current version of Mattr, which is to be expected as there were/are not many V11 Issuers available yet. As mentioned we for sure will be looking into this soon.

nklomp:
@genaris I create a new ticket #61 to track your issue. I also created a E2E test against the live Mattrlabs instance. 
The error message suggests that the Mattrlab Issuer is expecting an old payload. It also misses a required `nonce` value. I will bring it up with them and the test should become green as soon as they fix their issuer (or point out the error on our side, but I cannot see it ATM).

genaris:
> @genaris I created a new ticket #61 to track your issue. I also created a E2E test against the live Mattrlabs instance. The error message suggests that the Mattrlab Issuer is expecting an old payload. It also misses a required `nonce` value. I will bring it up with them and the test should become green as soon as they fix their issuer (or point out the error on our side, but I cannot see it ATM).

Thank you @nklomp for this analysis! Hope you can find soon a solution for this to make both compatible.

TimoGlastra:
Hey @nklomp, I'm not 100% up to date on all the changes that have been made to this library since I opened the PR, but I still think there's a lot of valuable changes in this PR, especially to unify the API and correct the usage of `type` (from the v8 spec) and `types` (from the VC @types) in newer versions.

Have you had the chance to take a closer look? Do you think all the issues have been adressed in main already and we can just update to 0.6.0?

nklomp:
Hy @TimoGlastra I will have a look at it later next week. There were some things in the PR I liked, but I also remember some things which I needed to look into because I was not sure about the approach.

PR #70, opened 2023-10-06
Feature/enable authorization code flow helpers

linasi:
- Removes checks so library could as well help with authorization code flow.
- Removed `flowType` parameter `OpenID4VCIClient`.
- Updated `acquirePushedAuthorizationRequestURI` method to return a complete authorization URL that can be opened by the client. Just like `createAuthorizationRequestUrl` does.
- Use `clientId` that was passed when constructing `OpenID4VCIClient` in subsequent calls.

### DIFF
--- a/packages/client/lib/OpenID4VCIClient.ts
+++ b/packages/client/lib/OpenID4VCIClient.ts
@@ -1,7 +1,6 @@
 import {
   AccessTokenResponse,
   Alg,
-  AuthorizationRequestV1_0_09,
   AuthzFlowType,
   CodeChallengeMethod,
   CredentialOfferPayloadV1_0_08,
@@ -38,7 +37,6 @@ interface AuthDetails {
 }
 
 interface AuthRequestOpts {
-  clientId: string;
   codeChallenge: string;
   codeChallengeMethod: CodeChallengeMethod;
   authorizationDetails?: AuthDetails | AuthDetails[];
@@ -47,7 +45,6 @@ interface AuthRequestOpts {
 }
 
 export class OpenID4VCIClient {
-  private readonly _flowType: AuthzFlowType;
   private readonly _credentialOffer: CredentialOfferRequestWithBaseUrl;
   private _clientId?: string;
   private _kid: string | undefined;
@@ -55,14 +52,7 @@ export class OpenID4VCIClient {
   private _endpointMetadata: EndpointMetadataResult | undefined;
   private _accessTokenResponse: AccessTokenResponse | undefined;
 
-  private constructor(
-    credentialOffer: CredentialOfferRequestWithBaseUrl,
-    flowType: AuthzFlowType,
-    kid?: string,
-    alg?: Alg | string,
-    clientId?: string,
-  ) {
-    this._flowType = flowType;
+  private constructor(credentialOffer: CredentialOfferRequestWithBaseUrl, kid?: string, alg?: Alg | string, clientId?: string) {
     this._credentialOffer = credentialOffer;
     this._kid = kid;
     this._alg = alg;
@@ -71,7 +61,6 @@ export class OpenID4VCIClient {
 
   public static async fromURI({
     uri,
-    flowType,
     kid,
     alg,
     retrieveServerMetadata,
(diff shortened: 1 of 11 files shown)


nklomp:
Cool thanks.   will have a look later today or probably tomorrow

packages/client/lib/OpenID4VCIClient.ts
```diff
@@ -133,36 +110,37 @@ export class OpenID4VCIClient {
...
+    if (!scope?.includes('openid')) {
+      scope = ['openid', scope].filter((s) => !!s).join(' ');
     }
 
-    //fixme: handle this for v11
-    const queryObj = {
+    const queryObj: { [key: string]: string } = {
       response_type: ResponseType.AUTH_CODE,
-      client_id: clientId,
+      client_id: this.clientId || '',
```
nklomp, line 119: Not entirely sure about using empty strings here

packages/client/lib/OpenID4VCIClient.ts
```diff
@@ -133,36 +110,37 @@ export class OpenID4VCIClient {
...
+
+    if (this.credentialOffer.issuerState) {
+      queryObj['issuer_state'] = this.credentialOffer.issuerState;
+    }
 
     return convertJsonToURI(queryObj, {
       baseUrl: this._endpointMetadata.authorization_endpoint,
-      uriTypeProperties: ['redirect_uri', 'scope', 'authorization_details'],
-      version: this.version(),
+      uriTypeProperties: ['redirect_uri', 'scope', 'authorization_details', 'issuer_state'],
```
nklomp, line 133: Will look later into whether leaving the version out doesn't have any side effects

packages/client/lib/OpenID4VCIClient.ts
```diff
@@ -183,21 +161,33 @@ export class OpenID4VCIClient {
...
+    if (!scope?.includes('openid')) {
+      scope = ['openid', scope].filter((s) => !!s).join(' ');
     }
 
-    //fixme: handle this for v11
-    const queryObj: AuthorizationRequestV1_0_09 = {
+    const queryObj: { [key: string]: string } = {
       response_type: ResponseType.AUTH_CODE,
-      client_id: clientId,
+      client_id: this.clientId || '',
```
nklomp, line 170: Not entirely sure about using empty strings here

nklomp (commented):
I will look over some of the initial comments I made later this weekend.

Overall looks great.

packages/client/lib/OpenID4VCIClient.ts
linasi: Yes, I agree. I have updated the logic to add `client_id` parameter only if it was provided during initialization.

packages/client/lib/OpenID4VCIClient.ts
linasi: Yes, I agree. I have updated the logic to add `client_id` parameter only if it was provided during initialization.

packages/client/lib/OpenID4VCIClient.ts
linasi: Based on the current `convertJsonToURI` implementation, excluding version param will do exactly what is needed in this case. Provided attributes have be added as query params. [else statement](https://github.com/Sphereon-Opensource/OID4VCI/blob/develop/packages/common/lib/functions/Encoding.ts#L36) does exactly that.

linasi:
Hi @nklomp, just wanted to let you know that parts you commented on have been addressed. Let me know if there is anything else that you find concerning.

packages/client/lib/OpenID4VCIClient.ts
nklomp: Yeah, the problem however is that the convertJsonToURI is called from many places and it for sure needs the logic to distinguish between V11 and higher in other areas. What I will do is add an additional param that takes precedence over the version. From the first look of it you need it from the `formPost` method. So will use that param there

nklomp:
@linasi Thanks for the PR. It is merged with a new param to make the encoding a bit more explicit

PR #74, opened 2023-10-23
fix: opts passed to getCredentialOfferEndpoint()

ragnika:


nklomp (approved):
Sorry for the late merge. Didn't notice this PR before.

Thanks for the submission

Issue #76, opened 2023-11-15
Separate generic format definitions and the format specific interfaces


TimoGlastra:
Currently the code is all a bit intertwined for different formats. This worked okay, but with different formats having different structures, it's become a bit hard to manage. This has mainly come up with adding support for the SD-JWT format, and that uses `credential_definition.type` , while JWT VC uses `types` (top level and plural), and JSON-LD uses `credentialDefinition.types` (camelCase and plural).

Should there have been more consistency between these types (definitely), but it shows that we shouldn't assume the format specific fields between formats will be similar, and thus to improve maintainability of this library, I think all format specific handling should be separated form the general protocol implementation.

Thus we have the base offer/request etc.. interfaces and handling, and then for each format you need to define the specific offer/request/issuer metadata interfaces, as well as some functions that can e.g. verify a request, or transform an offer of that specific format into a request. This could be an interface that is implemented with some common methods.

This would allow adding new formats without needing to hardcode them in this library, and would make it easier to see what is base OID4VCI code, and what is custom format specific logic

Thoughts on this?

nklomp:
Yes, makes sense at this point.

Maybe we can nicely combine this with #75 as there will be a new version which again has different behavior for the credential offers to metadata resolution/mapping

Issue #75, opened 2023-11-15
Remove support for versions below draft 11


TimoGlastra:
When developing on this library, over time it has lost some of the simplicity due to several formats and protocol versions being supported.

We have targeted Draft 8 before, but now are mainly working on Draft 11 and possibly newer drafts (maybe 19 or another one). To improve maintainability, I propose support for Draft 8 and 9 is removed. This will allow us to remove a lot of transformation code, and also simplify a lot of small issues currently encountered with this library.

Thoughts on removing support for drafts below 11? People still wanting to support these older drafts could stick to an older version of the library for now

nklomp:
Yeah agreed. For VP/SIOP we have to be a bit more careful, but for VCI it is not problem. We do have version 8 or 9 in the DIIP profile though 😱

TimoGlastra:
Hmm good point. But I thought we used draft 11 for DIIP. We showed interop for DIIP based on draft 11 (as that's the only one your issuer supports)

nklomp:
Yeah, somehow we managed to not specify the actual version for VCI in DIIP 😂 
Given the issuer has always been draft 11 anyway, as it was developer later then the client lib, let's just drop v8/v9 support altogether.

I think one discussion we need to have is whether we will wait for the next version in which the inline offers are removed and replaced by id's only or not and then only support that version

TimoGlastra:
It seems draft 12 has just been published: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html

Lots of changes here, with some notable ones the offer `credentials` being only strings now, and the credentials_supported being a map again, so we're back to what it was a couple of drafts ago it seems.

Do we just want to support that version? There's a lot of interop demonstrations currently based on draft 11, so I think dropping draft 11 will have a bigger impact. But with the vast amount of changes, it would be nice to just get rid of everything but the latest draft.

nklomp:
BTW small update on this one 

After the current SD-JWT work has landed, we will be updating this monorepo and also bring in the SIOP /OID4VP as modules. Reason is there is certain overlap and then having the modules next to one-another helps. In certain scenarios for instance doing a SIOP call from within a VCI context makes sense. That is of course next to having shared interfaces between the two of them.

We will also drop support for non-latest versions when starting that work as the specs seem to become a bit more stable. This means we will remove most of the version discovery, which should simplify the code drastically.
[...]

PR #78, opened 2023-11-16
feat: add sd-jwt support

TimoGlastra:
Adds support for `vc+sd-jwt` format as described in https://github.com/vcstuff/oid4vc-haip-sd-jwt-vc

Because the format specific for `vc+sd-jwt` are a bit different than the other formats, it exposed some issues with the types currently being used. I updated all interfaces, made the uniform interfaces match draft 11 fully, and moved some v8/9 specific types outside of the uniform/generic types.

Because the types are different, and it exposed some issues with the current way requests are being made, there's more format speicfic logic now inline. I've didn't want to make this PR bigger than it currently is, so I've created a separate issue #76, to explore that work.

All tests are passing except the e2e test with mattr launchpad, as it's currently down for maintenance. In addition I had to change some tests with vendors, as the expected types do not match with the spec. They all hardcode that the version is draft 8, but then they use a format that is somewhere in between draft 8 and 9. Mabye that's what those vendors implemented, or maybe it was just a mock that was created? I don't know, but I've corrected them now.

I think if we also address #75 at the same time we address #76, it can clean up a lot of format specific code that's now spread across the repository

### DIFF
--- a/packages/client/lib/CredentialRequestClient.ts
+++ b/packages/client/lib/CredentialRequestClient.ts
@@ -1,6 +1,7 @@
 import {
-  CredentialRequestV1_0_08,
   CredentialResponse,
+  getCredentialRequestForVersion,
+  getUniformFormat,
   OID4VCICredentialFormat,
   OpenId4VCIVersion,
   OpenIDResponse,
@@ -53,24 +54,7 @@ export class CredentialRequestClient {
   }
 
   public async acquireCredentialsUsingRequest(uniformRequest: UniformCredentialRequest): Promise<OpenIDResponse<CredentialResponse>> {
-    let request: CredentialRequestV1_0_08 | UniformCredentialRequest = uniformRequest;
-    if (!this.isV11OrHigher()) {
-      let format: string = uniformRequest.format;
-      if (format === 'jwt_vc_json') {
-        format = 'jwt_vc';
-      } else if (format === 'jwt_vc_json-ld') {
-        format = 'ldp_vc';
-      }
-
-      request = {
-        format,
-        proof: uniformRequest.proof,
-        type:
-          'types' in uniformRequest
-            ? uniformRequest.types.filter((t) => t !== 'VerifiableCredential')[0]
-            : uniformRequest.credential_definition.types[0],
-      } as CredentialRequestV1_0_08;
-    }
+    const request = getCredentialRequestForVersion(uniformRequest, this.version());
     const credentialEndpoint: string = this.credentialRequestOpts.credentialEndpoint;
     if (!isValidURL(credentialEndpoint)) {
       debug(`Invalid credential endpoint: ${credentialEndpoint}`);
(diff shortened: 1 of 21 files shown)


nklomp:
Cool thanks, will have a look later today

TimoGlastra:
FYI @nklomp, made a another PR that builds on top of this PR: https://github.com/TimoGlastra/OID4VCI/pull/1

But it needs a few tweaks before it's ready to be merged (need to get some input from my team on what the input looks like with SD-JWT VCs) as you need to provide some details about which fields can be disclosed, etc..

packages/client/lib/CredentialRequestClient.ts
```diff
@@ -92,45 +76,66 @@ export class CredentialRequestClient {
...
     const typesSelection =
       opts?.credentialTypes && (typeof opts.credentialTypes === 'string' || opts.credentialTypes.length > 0)
         ? opts.credentialTypes
         : this.credentialRequestOpts.credentialTypes;
     const types = Array.isArray(typesSelection) ? typesSelection : [typesSelection];
     if (types.length === 0) {
       throw Error(`Credential type(s) need to be provided`);
-    } else if (!this.isV11OrHigher() && types.length !== 1) {
+    }
+    // FIXME: this is mixing up the type (as id) from v8/v9 and the types (from the vc.type) from v11
```
nklomp, line 91: Could you elaborate a bit more on the FIXME. AFAIK current logic is following the different spec versions as it should (to be clear I didn't check explicitly when doing this review).

I haven't merged #54, because I believe it made some assumptions about 'inline credentials', which I wasn't too sure about.

IMO the approach should simply be to always resolve the offer to a credentials supported like object as it is also defined in the server metadata. So than we do not care whether it was an id, or some inline objects

nklomp (approved):
Had one comment about a fixme, but IMO we should solve that anyway by ensuring we drop support for older versions, so it is a bit of a mute discussion anyway

packages/client/lib/CredentialRequestClient.ts
nklomp: IMO merging current work and dropping support for some older versions is more important than this discussion and it will be mute anyway.

nklomp:
Let's keep the version discussion for #75. Per approved and to be merged.

packages/client/lib/CredentialRequestClient.ts
TimoGlastra: > ould you elaborate a bit more on the FIXME. AFAIK current logic is following the different spec versions as it should (to be clear I didn't check explicitly when doing this review).

So this is as I understand it, but between version 8 and 9/11 there was a change in what type means. Version 8 had a type which pointed to the credentials_supported. In draft 9/11 the `types` was added which referred to the actual values of the `type` in the VC (so an array and starts with `VerifiableCredential`). 

These were being used interchangeably in this library, and it resulted in some cases being handled a bit weirdly.  Because the getTypes method was adding the string values from an offer `credentials` together with the inline offer `types` property. And in this case the string values were pointers to the `id` property in `credentials_supported`, while the `types` were the types that would be included in the issued VC. I think in a lot of examples the `id` of a credential matched one of the `types` that would be put in the credential (e.g. `UniversityCredential` being used as the `credentials_supported.id`, as well as the second `type` in vc (`type: [VerifiableCredential, UniserityCredential]`). 

Hope this makes sense

PR #82, opened 2024-01-05
Feature/vdx 316  deferred credential support

nklomp:
Add deferred credential support

### DIFF
--- a/packages/client/lib/CredentialRequestClient.ts
+++ b/packages/client/lib/CredentialRequestClient.ts
@@ -1,7 +1,9 @@
 import {
+  acquireDeferredCredential,
   CredentialResponse,
   getCredentialRequestForVersion,
   getUniformFormat,
+  isDeferredCredentialResponse,
   OID4VCICredentialFormat,
   OpenId4VCIVersion,
   OpenIDResponse,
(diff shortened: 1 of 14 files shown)


packages/client/lib/CredentialRequestClient.ts
```diff
@@ -19,25 +21,57 @@ import { isValidURL, post } from './functions';
...
+  proofInput: ProofOfPossessionBuilder<DIDDoc> | ProofOfPossession,
+  opts: {
+    version: OpenId4VCIVersion;
+    cNonce?: string;
+  },
+) {
+  if ('proof_type' in proofInput) {
+    if (opts.cNonce) {
+      throw Error(`Cnonce param is only supported when using a Proof of Posession builder`);
+      //decodeJwt(proofInput.jwt).
```
maikel-maas, line 45: Remove commented-out code

packages/client/lib/CredentialRequestClient.ts
```diff
@@ -63,11 +97,40 @@ export class CredentialRequestClient {
...
+    },
+  ): Promise<OpenIDResponse<CredentialResponse>> {
+    const transactionId = response.transaction_id;
+    const bearerToken = response.acceptance_token ?? opts?.bearerToken;
+    const deferredCredentialEndpoint = this.getDeferredCredentialEndpoint();
+    if (!deferredCredentialEndpoint) {
+      throw Error(`No deferred credential endpoint supplied.`);
+    } else if (!bearerToken) {
+      throw Error(`No bearer token present and refresh for defered endpoint not supported yet`);
+      // todo updated bearer token with new c_nonce
```
maikel-maas, line 123: Is there a ticket for it?

maikel-maas (changes requested):
Left small remarks

packages/client/lib/CredentialRequestClient.ts
nklomp: No not needed, given it throws errors for this case. So if we really hit it (not likely soon), we know what it is about to begin with

packages/client/lib/CredentialRequestClient.ts
nklomp, line 45:
```suggestion
```

PR #83, opened 2024-01-11
fix(sd-jwt): cnf instead of kid

TimoGlastra:
small fix to use `cnf` object instead of `sub` property for binding the sd-jwt as defined in the sd-jwt vc spec.

Also adds support for `kid` (did) binding in addition to jwk binding for an sd-jwt


nklomp (approved):
LGTM

TimoGlastra:
updated with latest ssi-types

PR #89, opened 2024-02-13
feature/SDK-4

BtencateSphereon:


packages/client/lib/OpenID4VCIClient.ts
```diff
@@ -58,6 +61,11 @@ export class OpenID4VCIClient {
     credentialIssuer,
     pkce,
     authorizationRequest,
+    jwk,
+    endpointMetadata,
+    accessTokenResponse,
+    authorizationRequestOpts,
+    authorizationURL
   }: {
     credentialOffer?: CredentialOfferRequestWithBaseUrl;
```
nklomp, line 70: Why not use the new Interface here?

packages/client/lib/OpenID4VCIClient.ts
```diff
@@ -66,20 +74,34 @@ export class OpenID4VCIClient {
...
+      alg,
+      // TODO: We need to refactor this and always explicitly call createAuthorizationRequestUrl, so we can have a credential selection first and use the kid as a default for the client id
+      clientId: clientId ?? (credentialOffer && getClientIdFromCredentialOfferPayload(credentialOffer.credential_offer)) ?? kid?.split('#')[0],
+      pkce: { disabled: false, codeChallengeMethod: CodeChallengeMethod.S256, ...pkce },
+      jwk,
+      endpointMetadata,
+      accessTokenResponse,
+      authorizationURL
+    }
+    this._state = {
```
nklomp, line 100: Probably wise to make a comment why you need to split this up, given the sync method needs properties from the state.

packages/client/lib/OpenID4VCIClient.ts
```diff
@@ -66,20 +74,34 @@ export class OpenID4VCIClient {
...
+      ...this._state,
+      authorizationRequestOpts: authorizationRequestOpts ?? this.syncAuthorizationRequestOpts(authorizationRequest),
```
nklomp, line 102: Why not assign the Opts to this._state directly. We are still in the constructor

packages/client/lib/OpenID4VCIClient.ts
```diff
@@ -380,20 +408,24 @@ export class OpenID4VCIClient {
...
-        `Retrieving a credential from ${this._endpointMetadata
+        `Retrieving a credential from ${this._state.endpointMetadata
           ?.credential_endpoint} for issuer ${this.getIssuer()} failed as there was no success response body`,
       );
     }
     return response.successBody;
   }
 
+  public async exportState(): Promise<OpenID4VCIClientState> {
+    return this._state
```
nklomp, line 426: Probably wise to clone this, otherwise changes to the exported object would impact current client instance

nklomp (changes requested):
See remarks. Most important is the export clone remark

packages/client/lib/OpenID4VCIClient.ts
BtencateSphereon: it now returns a string

packages/client/lib/OpenID4VCIClient.ts
BtencateSphereon: made some changes, let me know if this is ok

packages/client/lib/OpenID4VCIClient.ts
BtencateSphereon: added a comment about running syncAuthorizationRequestOpts later

packages/client/lib/OpenID4VCIClient.ts
BtencateSphereon: because the options of the constructor are slightly different than the eventual state.

nklomp (approved):
LGTM, I will do the merge, since I need to look at an issue with CI anyway

Issue #88, opened 2024-02-12
Accept RSA Signing Algorithms


vincentkelleher:
Hi 👋 

In our goal of supporting OIDC4VCI/OIDC4VP protocols in the Gaia-X ecosystem we based ourselves on your `OID4VCI` lib which is very easy to get up to speed with thanks to the documentation and examples.

**Our Use Case**

Gaia-X Clearing Houses currently need an EV SSL certificate in order to be able to issue compliance verifiable credentials. Those certificates are mostly delivered with RSA signing algorithms and are costly so we wanted to try using those with your library to avoid having to ask members to buy new EV SSL certificates with `EdDSA`, `ES256` or `ES256K` signatures.

**The Issue**

We noticed that you implemented a closed list of algorithms with the `Alg` enum and that a check is done to make sure the credential request proof uses a supported algorithm. The following error is raised :

![image](https://github.com/Sphereon-Opensource/OID4VCI/assets/4272277/562b3a2d-66c8-4c5c-8870-1f85deab4de8)

I have tried adding the `PS256` algorithm we use to the `Alg` enum and everything works fine 👍 

**Thoughts**

As `jwtVerifyCallback` is the method that's depending on the signing algorithm and that this callback is defined by the developer using the library, hence its responsibility, wouldn't it be possible to remove this check and just leave it to the implementer ?

If not, would it be possible to add widely used algorithms to the `Alg` enum or even leave the implementing developer define it in the `VcIssuer` metadata ?

It would be my pleasure to exchange with you about this 😊

nklomp:
Hi @vincentkelleher Thanks for the nice words. Actually the OID4VC libs need an awful lot of work on the doc side, but given everything is changing below our feet, we only expect to do that over the next few months. Good to here you can find your bearing though.

Yeah you bring up some valid points. The primary reason is to protect the developer a bit from making mistakes by implementing an identifier that is not officially recognized, whilst also providing strong type support. At the same time I agree that we should not really be a barrier for new algorithms being implemented, since we do not make many assumptions about signing to begin with. Probably we should just extend the mechanism to also accept strings next to the enum. Or better yet move to a union type of string literal values as well as the string type. Then we have the best of both worlds. An IDE doing autocompletion/suggestions, but also allows having new Algos.

BTW happy to have short introduction call to see if and how we can help to begin with, given we also are working in certain Gaia-X related projects

nklomp:
[...]

PR #92, opened 2024-02-21
Open the signing algorithm list in the credential issuance process

vincentkelleher:
# Open the Signing Algorithm List

As a reference to #88, this Pull Request allows developers to use the given `Alg` list or a signing algorithm of their choice by just checking that the `alg` field is provided without verifying if its value is part of the `Alg` enum.

Some tests have been re-enabled, added and modified to have some coverage on this part.


sanderPostma:
LGTM

PR #97, opened 2024-03-11
fix: await session state updates

TimoGlastra:
Some state updates (set, delete) were not awaited. When updating the in memory session manager to one that uses persistance storage we were running into issues were we would close our database session (as the call to this library is finished), but then a write would still happen as the updates to the session were not awaited.

Were these awaits left out on purpose?


nklomp:
It was on purpose, to ensure any calling code would not be blocking on the session management itself. But for now let's indeed await. It needs some refactoring, simplification anyway soon.

nklomp (approved):

nklomp:
@TimoGlastra  0.10.1 has the fixes: https://github.com/Sphereon-Opensource/OID4VCI/releases/tag/v0.10.1

PR #101, opened 2024-03-13
fix: use seconds for all expires in values

TimoGlastra:
Seems there were some other places were milliseconds were used, but the specs describe to use seconds for `iat`, `exp`, `c_nonce_expires_in`, `interval` and `expires_in`

I modified some defaults, so this is technically a breaking change, but I think the current implementation is incorrect so it could also be seen as a fix I think.


nklomp (approved):
Hmzz, that is sloppy. Thanks.

Given we are still at 0.x, breaking changes are allowed ;)

TimoGlastra:
> Given we are still at 0.x, breaking changes are allowed ;)

That's why Credo has been in 0.x for years now 🙃

PR #104, opened 2024-03-17
fix: seconds to ms

TimoGlastra:
another fix related to milliseconds vs seconds

### DIFF
--- a/packages/issuer/lib/VcIssuer.ts
+++ b/packages/issuer/lib/VcIssuer.ts
@@ -519,7 +519,7 @@ export class VcIssuer<DIDDoc extends object> {
       }
       if (!iat) {
         throw new Error(IAT_ERROR)
-      } else if (iat > createdAt + tokenExpiresIn) {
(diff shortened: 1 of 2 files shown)


TimoGlastra:
@nklomp could you take a look at this PR?

nklomp:
Yeah, need to have a closer look into it. Because I wasn't entirely sure about it if I remember correctly.

nklomp:
To be clear it revolved more around also renaming the first part or not. I only quickly looked at it a few days ago from mobile phone

nklomp (approved):

nklomp:
Made a small change. The createdAt from the state manager is in milliseconds. The expiration arguments and JWT claims are in seconds. So the first change IMO was incorrect. Changed it to divide the createdAt by 1000 to convert it to seconds. Then add the tokenExpiresIn and compare that to iat

````typescript
iat > (createdAt/1000 + tokenExpiresIn)
````

PR #107, opened 2024-05-07
Feature/cwall 199

maikel-maas:


packages/client/README.md
```diff
@@ -51,6 +51,8 @@ client you can use to finish the pre-authorized code flows.
 This initiates the client using a URI obtained from the Issuer using a link (URL) or QR code typically. We are also
 already fetching the Server Metadata
 
+
+Using openid-initiate-issuance scheme
```
nklomp, line 55: As mentioned that is ancient. It is called credential offer for quite some time. So if you add documentation, don't use old terms

packages/client/lib/CredentialOfferClient.ts
```diff
@@ -27,22 +27,20 @@ export class CredentialOfferClient {
...
       credentialOffer = {
         credential_offer: credentialOfferPayload,
       };
     } else {
       credentialOffer = convertURIToJsonObject(uri, {
-        arrayTypeProperties: ['credentials'],
+        arrayTypeProperties: uri.includes('credential_offer_uri=') ? ['credential_offer_uri'] : ['credential_offer'],
         requiredProperties: uri.includes('credential_offer_uri=') ? ['credential_offer_uri'] : ['credential_offer'],
       }) as CredentialOfferV1_0_11;
-      if (credentialOffer?.credential_offer_uri === undefined && !credentialOffer?.credential_offer) {
```
nklomp, line 43: Why is this removed? There should never be a by value and by reference version together in a URL

packages/common/lib/functions/Encoding.ts
```diff
@@ -81,20 +89,47 @@ export function convertJsonToURI(
...
  *          - requiredProperties: the required properties
  *          - arrayTypeProperties: properties that can show up more that once
+ * @returns JSON object
  */
 export function convertURIToJsonObject(uri: string, opts?: DecodeURIAsJsonOpts): unknown {
   if (!uri || (opts?.requiredProperties && !opts.requiredProperties?.every((p) => uri.includes(p)))) {
     throw new Error(BAD_PARAMS);
   }
+
+  if (opts?.arrayTypeProperties && opts.arrayTypeProperties.includes('credential_offer_uri')) {
```
nklomp, line 104: Why does a generic function all of a sudden knows about credential_offers?

packages/common/lib/functions/Encoding.ts
```diff
@@ -81,20 +89,47 @@ export function convertJsonToURI(
...
+    return encodedCredentialOfferUri2Json({ uri })
+  }
+
+  if (opts?.arrayTypeProperties && opts.arrayTypeProperties.includes('credential_offer')) {
```
nklomp, line 108: Why does a generic function all of a sudden know about credential_offers ?

nklomp (changes requested):
See remarks

packages/client/README.md
maikel-maas: It was here already. Removed

packages/client/lib/CredentialOfferClient.ts
maikel-maas: Moved it back

nklomp: Yeah, but the question is. This was a rather self-explanatory error message with a really simple condition. That raises the question, why would you even decide to remove that without consulting?

packages/common/lib/functions/Encoding.ts
maikel-maas: Removed!

packages/common/lib/functions/Encoding.ts
maikel-maas: Removed!

packages/common/lib/functions/Encoding.ts
maikel-maas: Adding a simple '=' to the required properties fixed the issue

packages/common/lib/functions/Encoding.ts
maikel-maas: Adding a simple '=' to the required properties fixed the issue

packages/common/lib/functions/Encoding.ts
nklomp: Hmzz, no sure I like that approach, so please create a TODO. It should probably have been solved by not mixing the outer param with the inner param

packages/client/lib/CredentialOfferClient.ts
maikel-maas: I didn't introduce a ternary. The ternary was already there. I just added the '=', to make it skip the scheme and split the string at the credential offer query param.

packages/common/lib/functions/Encoding.ts
maikel-maas: The actual problem here is the input url. Usual url encoders encode the '='. In the examples provided in the specs of OID4VCI the '=' is not encoded. If we receive a fully encoded url it doesn't work.

nklomp (approved):

PR #108, opened 2024-05-08
feature/CWALL-174 based on Sander's work

sksadjad:
This is right now a work in progress. I've fixed the build issues and refactored code quite a bit. also added a few todos that needs discussing.
Right now fixing the test failures


packages/client/lib/AccessTokenClient.ts
```diff
@@ -29,7 +34,7 @@ export class AccessTokenClient {
     const { asOpts, pin, codeVerifier, code, redirectUri, metadata } = opts;
 
     const credentialOffer = opts.credentialOffer ? await assertedUniformCredentialOffer(opts.credentialOffer) : undefined;
-    const isPinRequired = credentialOffer && this.isPinRequiredValue(credentialOffer.credential_offer);
+    const pinMetadata: TxCodeAndPinRequired | undefined = credentialOffer && this.getPinMetadata(credentialOffer.credential_offer);
```
nklomp, line 37: Could we keep the isPinRequired as well?

packages/client/lib/AccessTokenClient.ts
```diff
@@ -132,28 +142,50 @@ export class AccessTokenClient {
...
+          // Create a regex for numeric input. If no length specified, allow any length of numeric input.
+          regex = length ? new RegExp(`^\\d{1,${length}}$`) : /^\d+$/;
+        } else if (input_mode === 'text') {
+          // Create a regex for text input. If no length specified, allow any length of alphanumeric input.
+          regex = length ? new RegExp(`^[a-zA-Z0-9]{1,${length}}$`) : /^[a-zA-Z0-9]+$/;
+        }
+      }
+
+      // Default regex for alphanumeric with no specific length limit if no input_mode is specified.
+      regex = regex || /^[a-zA-Z0-9]+$|^[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+$/;
```
nklomp, line 178: Probably makes more sense to move this before the if statements and let these overwrite it instead of the || yo are doing now

packages/client/lib/AuthorizationCodeClient.ts
```diff
@@ -17,18 +21,28 @@ import Debug from 'debug';
 
 const debug = Debug('sphereon:oid4vci');
 
+function filterSupportedCredentials(
+  credentialOffer: CredentialOfferPayloadV1_0_13,
+  credentialsSupported?: Record<string, CredentialConfigurationSupportedV1_0_13>,
+): CredentialConfigurationSupportedV1_0_13[] {
+  if (!credentialOffer.credential_configuration_ids || !credentialsSupported) {
+    return [];
```
nklomp, line 29: Are you sure this is an or. Looks more like it should be an and. If not, then this code is now tied to v13 by the looks of it. Probably wise to make a comment about that.
edit: I see that it gets called when v13 is detected, so please add the comment

packages/common/lib/types/Generic.types.ts
```diff
@@ -190,25 +240,25 @@ export interface CredentialRequestSdJwtVc extends CommonCredentialRequest {
 }
 
 export interface CommonCredentialResponse {
-  format: string;
+  // format: string;  TODO do we still need this for previous version support?
```
nklomp, line 243: Yeah, this was used for matching credential types previously. Not sure how tests can still pass if this is gone. Or has it been introduced elsewhere? Same for the other `format` properties that have been commented in this file

nklomp (changes requested):
Did a quick review. Some comments.

I think it is best to simply merge and fix issues from there

PR #111, opened 2024-06-12
Feature/cwall 227

sksadjad:
Added a function to add the token endpoint to the metadata

### DIFF
--- a/packages/issuer-rest/lib/oid4vci-api-functions.ts
+++ b/packages/issuer-rest/lib/oid4vci-api-functions.ts
@@ -2,7 +2,6 @@ import {
   ACCESS_TOKEN_ISSUER_REQUIRED_ERROR,
   adjustUrl,
   AuthorizationRequest,
-  CredentialIssuerMetadataOpts,
   CredentialOfferRESTRequest,
   CredentialRequestV1_0_11,
   determineGrantTypes,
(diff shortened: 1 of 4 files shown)


packages/issuer-rest/lib/oid4vci-api-functions.ts
```diff
@@ -379,9 +380,16 @@ export function pushedAuthorizationEndpoint<DIDDoc extends object>(
   })
 }
 
+function addTokenEndpoint(issuerMetadata: CredentialIssuerMetadataOpts) {
```
nklomp, line 383: setDefaultTokenEndpoint or something, as this method is not for adding token endpoints which the name suggests.

The real question is whether it should be done here, or in the VCIIssuer itself

nklomp (changes requested):
setDefaultTokenEndpoint or something, as this method is not for adding token endpoints which the name suggests.

The real question is whether it should be done here, or in the VCIIssuer itself

packages/issuer-rest/lib/oid4vci-api-functions.ts
sksadjad: renamed this to the name that you've suggested and moved it to VcIssuerBuilder

nklomp (approved):

PR #112, opened 2024-06-13
feature/SPRIND-14_based_on-draft13

sksadjad:
Ensure VCI client and server can work with plain JWKs not using DIDs
-  added support for `x5c`
- modified default value of `typ` to be `openid4vci-proof+jwt`

### DIFF
--- a/packages/client/lib/__tests__/CredentialRequestClient.spec.ts
+++ b/packages/client/lib/__tests__/CredentialRequestClient.spec.ts
@@ -31,7 +31,6 @@ import { getMockData } from './data/VciDataFixtures';
 const partialJWT = 'eyJhbGciOiJFUzI1NiJ9.eyJpc3MiOiJkaWQ6ZXhhbXBsZTplYmZlYjFmN';
 const partialJWT_withoutDid = 'eyJhbGciOiJFUzI1NiJ9.eyJpc3MiOiJlYmZlYjFmNzEyZWJjNmYxYzI3N';
 
-
 const jwt1_0_08: Jwt = {
   header: { alg: Alg.ES256, kid: 'did:example:ebfeb1f712ebc6f1c276e12ec21/keys/1', typ: 'jwt' },
   payload: { iss: 'sphereon:wallet', nonce: 'tZignsnFbp', jti: 'tZignsnFbp223', aud: IDENTIPROOF_ISSUER_URL },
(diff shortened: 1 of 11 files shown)


packages/client/lib/functions/ProofUtil.ts
```diff
@@ -84,12 +85,13 @@ const createJWT = (jwtProps?: JwtProps, existingJwt?: Jwt): Jwt => {
...
   const jti = getJwtProperty<string>('jti', false, jwtProps?.jti, existingJwt?.payload?.jti);
-  const typ = getJwtProperty<string>('typ', true, jwtProps?.typ, existingJwt?.header?.typ, 'jwt');
+  const typ = getJwtProperty<string>('typ', true, jwtProps?.typ, existingJwt?.header?.typ, 'openid4vci-proof+jwt');
   const nonce = getJwtProperty<string>('nonce', false, jwtProps?.nonce, existingJwt?.payload?.nonce); // Officially this is required, but some implementations don't have it
   // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
   const alg = getJwtProperty<string>('alg', false, jwtProps?.alg, existingJwt?.header?.alg, 'ES256')!;
-  const kid = getJwtProperty<string>('kid', false, jwtProps?.kid, existingJwt?.header?.kid);
+  const kid = jwtProps?.jwk ? undefined : getJwtProperty<string>('kid', false, jwtProps?.kid, existingJwt?.header?.kid);
   const jwk = getJwtProperty<BaseJWK>('jwk', false, jwtProps?.jwk, existingJwt?.header?.jwk);
+  const x5c = jwtProps?.x5c;
```
nklomp, line 94: Why isn't this following the same pattern as the other claims?

packages/client/lib/functions/ProofUtil.ts
```diff
@@ -104,8 +106,9 @@ const createJWT = (jwtProps?: JwtProps, existingJwt?: Jwt): Jwt => {
   const jwtHeader: JWTHeader = {
     typ,
     alg,
-    kid,
-    jwk,
+    ...(kid ? { kid } : {}),
```
nklomp, line 109: ...(kid && {kid}) etc

nklomp (changes requested):
Okay this is a start, but probably way more is needed to make the library in a state where it doesn't assume DIDs

packages/client/lib/__tests__/CredentialRequestClient.spec.ts
```diff
@@ -136,6 +186,28 @@ describe('Credential Request Client ', () => {
       Error(URL_NOT_VALID),
     );
   });
+
+  it('should fail with invalid url', async () => {
+    const credReqClient = CredentialRequestClientBuilder.fromCredentialOfferRequest({ request: INITIATION_TEST })
+      .withCredentialEndpoint('httpsf://oidc4vci.demo.spruceid.com/credential')
```
cre8, line 192: is `httpsf` correct or is the f just a typo?

nklomp: On purpose given the name of the test, but indeed could be more clear

PR #115, opened 2024-06-20
SPRIND-14

sksadjad:
added x5c support and made sure that we support request-responses without dids

### DIFF
--- a/packages/client/lib/CredentialRequestClientBuilder.ts
+++ b/packages/client/lib/CredentialRequestClientBuilder.ts
@@ -1,33 +1,16 @@
 import {
-  AccessTokenResponse,
-  CredentialIssuerMetadataV1_0_13,
-  CredentialOfferPayloadV1_0_13,
   CredentialOfferRequestWithBaseUrl,
   determineSpecVersionFromOffer,
   EndpointMetadata,
-  ExperimentalSubjectIssuance,
-  getIssuerFromCredentialOfferPayload,
-  OID4VCICredentialFormat,
   OpenId4VCIVersion,
   UniformCredentialOfferRequest,
 } from '@sphereon/oid4vci-common';
-import { CredentialFormat } from '@sphereon/ssi-types';
 
 import { CredentialOfferClient } from './CredentialOfferClient';
-import { CredentialRequestClient } from './CredentialRequestClient';
+import { CredentialRequestClientBuilderV1_0_11 } from './CredentialRequestClientBuilderV1_0_11';
+import { CredentialRequestClientBuilderV1_0_13 } from './CredentialRequestClientBuilderV1_0_13';
 
 export class CredentialRequestClientBuilder {
-  credentialEndpoint?: string;
-  deferredCredentialEndpoint?: string;
-  deferredCredentialAwait = false;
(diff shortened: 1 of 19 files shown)


packages/client/lib/__tests__/IT.spec.ts
```diff
@@ -222,164 +287,36 @@ describe('OID4VCI-Client should', () => {
...
-        proofInput: proof,
-        credentialTypes: credentialOffer.original_credential_offer.credential_configuration_ids[0],
-      });
-      expect(credResponse.successBody?.credential).toEqual(mockedVC);
-    },
-    UNIT_TEST_TIMEOUT,
-  );
-});
-
-describe('OIDVCI-Client for v1_0_13 should', () => {
```
nklomp, line 292: Why is this removed?

packages/client/lib/__tests__/IT.spec.ts
```diff
@@ -222,164 +287,36 @@ describe('OID4VCI-Client should', () => {
...
-    nock(ISSUER_URL)
-    .post(/credential/)
-    .reply(200, {
-      format: 'jwt-vc',
-      credential: mockedVC,
-    });
-  }*/
-
-  it('should successfully resolve the credential offer URI', async () => {
-    const uri = 'https://mijnkvk.acc.credenco.com/openid4vc/credentialOffer?id=32fc4ebf-9e31-4149-9877-e3c0b602d559';
```
nklomp, line 339: Why is this removed?

nklomp (changes requested):
Some questions on removed tests

packages/client/lib/__tests__/IT.spec.ts
sksadjad: added back. not sure why it's not showing it here

packages/client/lib/__tests__/IT.spec.ts
sksadjad: added back

packages/client/lib/CredentialRequestClientBuilder.ts
```diff
@@ -40,24 +23,31 @@ export class CredentialRequestClientBuilder {
...
+      return CredentialRequestClientBuilderV1_0_13.fromCredentialIssuer({
+        credentialIssuer,
+        metadata,
+        version,
+        credentialIdentifier,
+        credentialTypes,
+      });
+    } else {
+      if (!credentialTypes || credentialTypes.length === 0) {
+        throw new Error('CredentialTypes must be provided for v1_0_11');
```
nklomp, line 58: Why doesn't the v11 builder take care of this?

nklomp (commented):
See remark about code outside of the v11 builder doing a check, that probably the v11 builder should do itself

packages/client/lib/CredentialRequestClientBuilder.ts
sksadjad: this is because in the `redentialRequestClientBuilderV1_0_11.fromCredentialIssuer` the `credentialTypes` param is required and in the draft13 it is optional. so I think it's better for it to be optional in the general `fromCredentialIssuer` function and we only check if it's passing a value or not.

PR #118, opened 2024-06-26
fix: update tx_code check

cre8:
add correct check for tx code check, since it got replaced in draft 13

PR #121, opened 2024-07-04
fix: sdjwtvc decode hasher dependency

auer-martin:
This PR fixes two small problems.

In the VcIssuer file, the issuer is determined by decoding the credential into a standardized format with an issuer field.
Decoding the issuer requires a hash implementation for compact sdjwt's. To work around this, I made the issuer an argument.
The second problem concerns the implementation of token validation in combination with a PIN and compatibility with v11 and v13.


nklomp:
Hmzz, probably some tests will fail as this one has overlap with #118

nklomp:
@auer-martin Suggest to leave out the tx_code fix. We will then address that in #118

auer-martin:
@nklomp done

nklomp (approved):
LGTM, thx

PR #128, opened 2024-07-16
fix: set userPin back to fix current implementation

cre8:
Adds userPin back to the CredentialOfferSession

@nklomp this is just a fix to keep the current library working again. It would be better to replace userPin in all libraries. However this would break existing APIs


cre8:
mmh, when I tried to run the test on the develop branch I am getting 1 failed, 2 skipped and 30 passed. Could it be a problem because I am using pnpm v9 instead of 8?

PR #129, opened 2024-07-19
fix: oid4vci draft 13 typing

TimoGlastra:
A lot of the typing for draft 13 was not according to the spec. I also separated some type definitions per format so we have better type completion. We noticed when integrating draft 13 vs draft 11 in Credo that with draft 13 we lost a lost of format specific typing, which made using the library a lot easier. This aims to bring it back.

With the updates of the types I also discovered some bugs related to transformations, and which keys were used (e.g. `types` for v11 were still used for v13 even though it should be `type` now. As well as request in v13 needs to use `credential_definition` for jwt vc json as well now, which in v11 was still top-level).

### DIFF
--- a/packages/client/lib/AuthorizationCodeClient.ts
+++ b/packages/client/lib/AuthorizationCodeClient.ts
@@ -5,14 +5,16 @@ import {
   convertJsonToURI,
   CreateRequestObjectMode,
   CredentialConfigurationSupportedV1_0_13,
+  CredentialDefinitionJwtVcJsonLdAndLdpVcV1_0_13,
+  CredentialDefinitionJwtVcJsonV1_0_13,
   CredentialOfferPayloadV1_0_13,
   CredentialOfferRequestWithBaseUrl,
   determineSpecVersionFromOffer,
   EndpointMetadataResultV1_0_13,
   formPost,
+  isW3cCredentialSupported,
   JsonURIMode,
   Jwt,
-  OID4VCICredentialFormat,
   OpenId4VCIVersion,
   PARMode,
   PKCEOpts,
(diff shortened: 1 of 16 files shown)


packages/client/lib/AuthorizationCodeClient.ts
```diff
@@ -127,42 +132,42 @@ export const createAuthorizationRequestUrl = async ({
...
-      if (!credential_configuration_id) {
-        format = cred.format;
-      }
       if (!credential_configuration_id && !cred.format) {
         throw Error('format is required in authorization details');
       }
 
-      const meta: any = {};
-      const credential_definition = cred.credential_definition;
-      if (credential_definition?.type && !format) {
```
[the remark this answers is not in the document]
nklomp: Okay, upon further inspection I was too quick. It is doing what is supposed to be happening

packages/common/lib/functions/TypeConversionUtils.ts
```diff
@@ -10,13 +25,19 @@ export const getNumberOrUndefined = (input?: string): number | undefined => {
...
+    | CredentialDefinitionJwtVcJsonV1_0_13
+    | JsonLdIssuerCredentialDefinition
+    | string,
 ): string[] | undefined {
   if (subject === undefined) {
     return undefined;
   } else if (typeof subject === 'string') {
     return [subject];
-  } else if ('credential_definition' in subject && subject.credential_definition) {
+  } else if ('format' in subject && subject.format !== 'vc+sd-jwt' && 'credential_definition' in subject) {
```
nklomp, line 40: I am not entirely sure whether format would always be present for this expression branch given the amount of types this function accepts. So probably better to change the logic around to search for credential_definition as long as there is no format or format not equal to sd-jwt. So something like !format or format !== sd-jwt'

packages/common/lib/functions/CredentialRequestUtil.ts
```diff
@@ -11,21 +11,31 @@ import { getFormatForVersion } from './FormatUtils';
 
 export function getTypesFromRequest(credentialRequest: CredentialRequest, opts?: { filterVerifiableCredential: boolean }) {
   let types: string[] = [];
+  console.log(credentialRequest);
```
nklomp, line 14: Remove console.log

nklomp (changes requested):
2 small remakrs. Rest LGTM. Okay to merge after being addressed

packages/common/lib/functions/TypeConversionUtils.ts
TimoGlastra: Hmm all the types that are supported for `subject` in the function parameter, either have no `credential_definition` defined, or will always have both `format` and `credential_definition` defined. 

But changed it now to not check format. This method is supposed to catch all, so being permissive is probably good.

packages/client/lib/AuthorizationCodeClient.ts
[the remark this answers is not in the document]
TimoGlastra: Yeah it's a bit complex how this works in the spec ..., but it seems to be correct

TimoGlastra:
Resolved comments!

nklomp (approved):

PR #131, opened 2024-07-25
feat: dpop support

auer-martin:
This PR adds support for [dPoP](https://www.rfc-editor.org/rfc/rfc9449.html).
DPoP support itself did not require many changes. Most of the changes made in this PR are refactorings, e.g., JwtIssuer and JwtVerifier to common so they can be reused ....sorry for that. 
Changes made for actual dpop support can be found in this [commit](https://github.com/Sphereon-Opensource/OID4VC/commit/92026678c745b770957f5bae290ae7b456601fd2).
I open the PR as a draft for now, as I plan to go over the spec one last time in the next couple of days and add some tests.


packages/siop-oid4vp/lib/request-object/Payload.ts
```diff
@@ -1,4 +1,4 @@
-import { v4 as uuidv4 } from 'uuid'
+import { uuidv4 } from '@sphereon/oid4vci-common'
```
nklomp, line 1: This and the JWK thumbprint changes do result in the SIOP lib depending on the VCI lib. Right now the oid4vci-common lib still is very much vci only. Probably it would be wise to create a `common` lib, that vci-common and siop can depend upon, where we place the real common stuf

nklomp (changes requested):
Overall looking very good. Some small suggestions, comments.

I do think we probably need to create a real `common` module. Because with the changes proposed (which are good) now the SIOP package will start depending on the vci-common package. The latter really is about VCI ATM, so probably better to start a real common package, which will also server as a nice module for the refactor happening soon

packages/issuer-rest/lib/IssuerTokenEndpoint.ts
```diff
@@ -37,16 +40,51 @@ export const handleTokenRequest = <T extends object>({
...
+    if (request.headers.dpop) {
+      if (!dPoPVerifyJwtCallback) {
+        return sendErrorResponse(response, 400, {
+          error: TokenErrorResponse.invalid_request,
+          error_description: 'DPOP is not supported',
+        })
+      }
+
+      try {
+        const fullUrl = request.protocol + '://' + request.get('host') + request.originalUrl
```
TimoGlastra, line 60: Host is not always defined or the same url. Host can be 0.0.0.0 while you use a public url. I think this needs to be configured somewhere (i've bad issues with this in credo when deploying to a https server url)

packages/issuer-rest/lib/IssuerTokenEndpoint.ts
```diff
@@ -37,16 +40,51 @@ export const handleTokenRequest = <T extends object>({
...
 
+    if (request.headers.authorization && request.headers.authorization.startsWith('DPoP ') && !request.headers.DPoP) {
+      return sendErrorResponse(response, 400, {
+        error: TokenErrorResponse.invalid_request,
+        error_description: 'DPoP header is required',
+      })
+    }
+
+    let dPoPJwk: JWK | undefined
+    if (request.headers.dpop) {
```
TimoGlastra, line 51: Is there a way to require dpop here?

TimoGlastra (commented):
Very nice martin!

packages/common/lib/functions/DPoP.ts
```diff
@@ -0,0 +1,218 @@
...
+  dPoPSigningAlgValuesSupported?: string[];
+}
+
+export type CreateDPoPClientOptions = CreateDPoPOptions<Omit<CreateDPoPJwtPayloadProps, 'htm' | 'htu'>>;
+
+export async function createDPoP(options: CreateDPoPOptions): Promise<string> {
+  const { createJwtCallback, jwtIssuer, jwtPayloadProps, dPoPSigningAlgValuesSupported } = options;
+
+  if (jwtPayloadProps.accessToken && (jwtPayloadProps.accessToken?.startsWith('DPoP ') || jwtPayloadProps.accessToken?.startsWith('Bearer '))) {
+    throw new Error('expected accessToken without scheme');
```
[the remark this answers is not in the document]
auer-martin: I called it a scheme because the specification calls it a scheme. I personally like that because it's easier to find things in the specifications later. So, I'm not sure. Should I change it?

packages/client/lib/AccessTokenClientV1_0_11.ts
```diff
@@ -91,10 +96,18 @@ export class AccessTokenClientV1_0_11 {
           : undefined,
     });
 
-    return this.sendAuthCode(requestTokenURL, accessTokenRequest);
+    let dPoP: string | undefined;
+    if (createDPoPOptions?.dPoPSigningAlgValuesSupported && createDPoPOptions.dPoPSigningAlgValuesSupported.length > 0) {
+      const htu = requestTokenURL.split('?')[0].split('#')[0];
```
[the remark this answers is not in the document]
auer-martin: This is being implemented clientside and only takes the tokenUrl from the metadata, so I think the URL should be fine. What checks would you implement?

packages/client/lib/AccessTokenClient.ts
```diff
@@ -87,10 +92,17 @@ export class AccessTokenClient {
           : undefined,
     });
 
-    return this.sendAuthCode(requestTokenURL, accessTokenRequest);
+    let dPoP: string | undefined;
+    if (createDPoPOptions?.dPoPSigningAlgValuesSupported && createDPoPOptions?.dPoPSigningAlgValuesSupported.length > 0) {
+      const htu = requestTokenURL.split('?')[0].split('#')[0];
```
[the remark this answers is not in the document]
auer-martin: htu: The HTTP target URI ([Section 7.1](https://rfc-editor.org/rfc/rfc9110#section-7.1) of [[RFC9110](https://www.rfc-editor.org/rfc/rfc9449.html#RFC9110)]) of the request to which the JWT is attached, without query and fragment parts.
This is what the spec says. IMO this should be sufficient. What do you think?

packages/client/lib/AccessTokenClientV1_0_11.ts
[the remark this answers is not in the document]
nklomp: The split first strips of the query and then strips of the fragment. I am not sure what the preconditions are of the requestTokenURL, but I would expect a bit of expressions checking whether your preconditions is actually satisfied.

Anyway since you are doing the same splits somewhere else I would create a function at least.

packages/issuer-rest/lib/IssuerTokenEndpoint.ts
auer-martin: Hmmm, I am not sure about it. I don't mind adding it as a parameter to the handleTokenRequest, but it looks weird to have it pass as a parameter. 
> @nklomp You also had a question about this. 
> Please doublecheck whether request.protocol isn't already including the trailing :
> https://github.com/expressjs/express/issues/4697#issue-994058715
> https://stackoverflow.com/questions/10183291/how-to-get-the-full-url-in-express/10185427#10185427

For now, I changed the implementation so that the access token endpoint URL can be passed; if passed as an argument, we use it. Otherwise, we fall back to the above behavior. We can also enforce passing it. What do you think?

packages/common/lib/functions/JwtIssuer.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export interface JwtIssuerJwk extends JwtIssuerBase {
+  method: 'jwk';
+  alg: SigningAlgo;
+  jwk: JsonWebKey;
+}
+
+export interface JwtIssuerCustom extends JwtIssuerBase {
+  method: 'custom';
```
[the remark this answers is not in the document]
auer-martin: no custom will not do anything

packages/client/lib/CredentialRequestClient.ts
```diff
@@ -89,6 +91,7 @@ export class CredentialRequestClient {
     context?: string[];
     format?: CredentialFormat | OID4VCICredentialFormat;
     subjectIssuance?: ExperimentalSubjectIssuance;
+    createDPoPOptions?: CreateDPoPClientOptions;
```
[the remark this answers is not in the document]
auer-martin: IMO not the dpop to me is the proof itself not to options for creating it.

packages/callback-example/package.json
```diff
@@ -10,6 +10,7 @@
     "build:clean": "tsc --build --clean && tsc --build"
   },
   "dependencies": {
+    "@sphereon/common": "workspace:*",
```
nklomp, line 13: Sorry I probably should have been more clear. I meant `@sphereon/oid4vc-common`.

auer-martin: done

nklomp: Okay let me go over the rest

packages/client/lib/AccessTokenClientV1_0_11.ts
```diff
@@ -91,10 +97,33 @@ export class AccessTokenClientV1_0_11 {
...
+      createDPoPOpts.jwtPayloadProps.nonce = retryWithNonce.dpopNonce;
+
+      dPoP = await createDPoP(getCreateDPoPOptions(createDPoPOpts, requestTokenURL));
+      response = await this.sendAuthCode(requestTokenURL, accessTokenRequest, dPoP ? { headers: { dpop: dPoP } } : undefined);
+      const successDPoPNonce = response.origResponse.headers.get('DPoP-Nonce');
+
+      nextDPoPNonce = successDPoPNonce ?? retryWithNonce.dpopNonce;
+    }
+
+    if (response.successBody && createDPoPOpts && createDPoPOpts && response.successBody.token_type !== 'DPoP') {
```
nklomp, line 117: This for sure will ensures the DPoPOpts is present :D

nklomp, line 117:
```suggestion
    if (response.successBody && createDPoPOpts && response.successBody.token_type !== 'DPoP') {
```

packages/client/lib/functions/dpopUtil.ts
```diff
@@ -0,0 +1,33 @@
...
+import { OpenIDResponse } from 'oid4vci-common';
+
+export function dPoPShouldRetryRequestWithNonce(response: OpenIDResponse<unknown, unknown>) {
+  if (response.errorBody && response.errorBody.error === dpopTokenRequestNonceError) {
+    const dPoPNonce = response.errorBody.headers.get('DPoP-Nonce');
+    if (!dPoPNonce) {
+      throw new Error('The DPoP nonce was not returned');
+    }
+
+    return { ok: true, dpopNonce: dPoPNonce } as const;
```
nklomp, line 11: Can we create a small type for this? The ok is not immediately clear. I think having the http  status code instead of ok would be nicer. The 401 is because we handle it nicely I guess, but it could also happen with a 400. So probably just using the http status is quite useful

packages/client/lib/CredentialRequestClientV1_0_11.ts
```diff
@@ -83,17 +88,45 @@ export class CredentialRequestClientV1_0_11 {
...
+    let response = (await post(credentialEndpoint, JSON.stringify(request), {
+      bearerToken: requestToken,
+      customHeaders: { ...(createDPoPOpts && { dpop: dPoP }) },
+    })) as OpenIDResponse<CredentialResponse> & {
       access_token: string;
     };
+
+    let nextDPoPNonce = createDPoPOpts?.jwtPayloadProps.nonce;
+    const retryWithNonce = dPoPShouldRetryResourceRequestWithNonce(response);
+    if (retryWithNonce.ok && createDPoPOpts) {
```
nklomp, line 103: What if retryWithNonce.false && createDPopOpts?

packages/client/lib/AccessTokenClient.ts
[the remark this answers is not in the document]
nklomp: I guess it is

packages/common/lib/functions/DPoP.ts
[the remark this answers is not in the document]
nklomp: No if the spec calls it like that, let's follow it

packages/client/lib/CredentialRequestClient.ts
```diff
@@ -119,9 +124,33 @@ export class CredentialRequestClient {
...
+    let response = (await post(credentialEndpoint, JSON.stringify(request), {
+      bearerToken: requestToken,
+      customHeaders: { ...(dPoP && { dpop: dPoP }) },
+    })) as OpenIDResponse<CredentialResponse> & {
       access_token: string;
     };
+
+    let nextDPoPNonce = createDPoPOpts?.jwtPayloadProps.nonce;
+    const retryWithNonce = dPoPShouldRetryResourceRequestWithNonce(response);
+    if (retryWithNonce.ok && createDPoPOpts) {
```
nklomp, line 139: What if retryWithNonce.false && createDPopOpts?

packages/client/lib/AccessTokenClientV1_0_11.ts
```diff
@@ -91,10 +97,33 @@ export class AccessTokenClientV1_0_11 {
...
 
-    return this.sendAuthCode(requestTokenURL, accessTokenRequest);
+    const useDpop = createDPoPOpts?.dPoPSigningAlgValuesSupported && createDPoPOpts.dPoPSigningAlgValuesSupported.length > 0;
+    let dPoP = useDpop ? await createDPoP(getCreateDPoPOptions(createDPoPOpts, requestTokenURL)) : undefined;
+
+    let response = await this.sendAuthCode(requestTokenURL, accessTokenRequest, dPoP ? { headers: { dpop: dPoP } } : undefined);
+
+    let nextDPoPNonce = createDPoPOpts?.jwtPayloadProps.nonce;
+    const retryWithNonce = dPoPShouldRetryRequestWithNonce(response);
+    if (retryWithNonce.ok && createDPoPOpts) {
```
nklomp, line 107: What if retryWithNonce.false && createDPopOpts?

packages/common/lib/dpop/DPoP.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  // Validate nonce if provided
+  if ((options.expectedNonce && dPoPPayload.nonce !== options.expectedNonce) || (!options.expectedNonce && dPoPPayload.nonce)) {
+    throw new Error('invalid_dpop_proof. Nonce mismatch');
+  }
+
+  // Validate iat claim
+  const { nowSkewedPast, nowSkewedFuture } = getNowSkewed(options.now);
+  if (
+    // iat claim is to far in the future
```
nklomp, line 173:
```suggestion
    // iat claim is too far in the future
```

packages/common/lib/dpop/DPoP.ts
```diff
@@ -0,0 +1,242 @@
...
+    nowSkewedPast - (options.maxIatAgeInSeconds ?? 300) > dPoPPayload.iat ||
```
nklomp, line 174: I would make the defaults a bit smaller. Systems being of 5 minutes would be very bad. Suggest to use 60

packages/common/lib/index.ts
```diff
@@ -1,9 +1,9 @@
 import { Loggers } from '@sphereon/ssi-types';
 
 export const VCI_LOGGERS = Loggers.DEFAULT;
-export const VCI_LOG_COMMON = VCI_LOGGERS.get('sphereon:oid4vci:common');
+export const VCI_LOG_COMMON = VCI_LOGGERS.get('sphereon:common');
```
nklomp, line 4: Please leave 'oid4vc' in there. We use these namespaces in our DEBUG logging to be able to selectively log certain namespaces. Now it is being moved up to the sphereon namspace

nklomp, line 4:
```suggestion
export const VCI_LOG_COMMON = VCI_LOGGERS.get('sphereon:oid4vc:common');
```

packages/common/lib/jwt/jwtUtils.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/**
+ * The maximum allowed clock skew time in seconds. If an time based validation
+ * is performed against current time (`now`), the validation can be of by the skew
+ * time.
+ *
+ * See https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.5
+ */
+const DEFAULT_SKEW_TIME = 300;
```
nklomp, line 26: See other remark. Suggest to lower this to 60

packages/issuer-rest/lib/IssuerTokenEndpoint.ts
```diff
@@ -37,16 +44,58 @@ export const handleTokenRequest = <T extends object>({
...
+    if (requireDPoP && !request.headers.dpop) {
+      return sendErrorResponse(response, 400, {
+        error: TokenErrorResponse.invalid_request,
+        error_description: 'DPoP is required for requesting access tokens',
+      })
+    }
+
+    if (request.headers.dpop) {
+      if (!dPoPVerifyJwtCallback) {
+        return sendErrorResponse(response, 400, {
```
nklomp, line 64: Do we want to log something or make the error a bit more clear. There is a distinction in not enabling DPOP and not having provided the verify callback. In other words, for anyone configuring the issuer, this might be a confusing error

nklomp (changes requested):
Thanks @auer-martin . This probably became a way bigger PR than imagined. 😂 

A few small remarks. My only bigger issue remaining is the retryDPoP result object. I think we should create a small type for it and I am also not sure about the current handling of these methods when the dpop create options are set and the dpop methods return ok: false.

packages/client/lib/CredentialRequestClient.ts
auer-martin: That's valid. For example, the request already succeeded without the nonce, in which case we don't need to retry it.

I agree that createDPoPOpts is a bit annoying here. I included it because otherwise, we need to throw an Error in case retryWithNonce. OK === true and createDPoPOpts is not set, which is a combination that is not possible. Would you prefer to throw?

packages/client/lib/functions/dpopUtil.ts
auer-martin: I created a type. Hmmmmm, Not sure about the HTTP status return. Because then, at a higher level, you need to know which status codes trigger the retry.

The 401 is required for Resource Requests. https://www.rfc-editor.org/rfc/rfc9449.html#name-resource-server-provided-no

packages/issuer-rest/lib/IssuerTokenEndpoint.ts
auer-martin: I made the drop input params an object and logged an error for the issuer controller.

auer-martin:
> Thanks @auer-martin . This probably became a way bigger PR than imagined. 😂

It was indeed a lot more work and took much longer than originally expected 🤣.

TimoGlastra (approved):
Tested the PR in our wallet and succesfully received credentials from the Funke PID issuer.

TimoGlastra:
@nklomp leaving final review / approval to you

nklomp (approved):
LGTM. Will merge

PR #133, opened 2024-07-30
fix: scope and par fixes

TimoGlastra:
This PR makes two changes:
- prioritise authorization server metadata over issuer metadata for PAR requirements (Funke issuer adds it in authorization server metadata)
- do not include `openid` scope if not included. I can't find in the specs where this is a requirement, and the funke PID issuer will not grant this scope. I think it makes more sense to let the user pass the scope `openid` if it wants this scope.

### DIFF
--- a/packages/client/lib/AuthorizationCodeClient.ts
+++ b/packages/client/lib/AuthorizationCodeClient.ts
@@ -113,13 +113,16 @@ export const createAuthorizationRequestUrl = async ({
   const { redirectUri, requestObjectOpts = { requestObjectMode: CreateRequestObjectMode.NONE } } = authorizationRequest;
   const client_id = clientId ?? authorizationRequest.clientId;
 
-  let { scope, authorizationDetails } = authorizationRequest;
-  const parMode = endpointMetadata?.credentialIssuerMetadata?.require_pushed_authorization_requests
+  // Authorization server metadata takes precedence
+  const authorizationMetadata = endpointMetadata.authorizationServerMetadata ?? endpointMetadata.credentialIssuerMetadata
+
+  let { authorizationDetails } = authorizationRequest;
+  const parMode = authorizationMetadata?.require_pushed_authorization_requests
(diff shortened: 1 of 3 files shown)


packages/client/lib/AuthorizationCodeClient.ts
```diff
@@ -113,13 +113,16 @@ export const createAuthorizationRequestUrl = async ({
   const { redirectUri, requestObjectOpts = { requestObjectMode: CreateRequestObjectMode.NONE } } = authorizationRequest;
   const client_id = clientId ?? authorizationRequest.clientId;
 
-  let { scope, authorizationDetails } = authorizationRequest;
-  const parMode = endpointMetadata?.credentialIssuerMetadata?.require_pushed_authorization_requests
+  // Authorization server metadata takes precedence
+  const authorizationMetadata = endpointMetadata.authorizationServerMetadata ?? endpointMetadata.credentialIssuerMetadata
+
+  let { authorizationDetails } = authorizationRequest;
```
nklomp, line 119: Is there a specific reason why you removed `scope` here? You are now using `authorizationRequest.scope` a few times below. Nothing big, just easier to read

nklomp (commented):
Just a small remark, not really needed

packages/client/lib/AuthorizationCodeClient.ts
TimoGlastra: It was complaining about let vs const. So chose this over 2 destructurings or an eslint ignore. Happy to change it though

nklomp: change to const would have fixed it right :P

TimoGlastra: No because authorizationDetails is reassigned.

So then it should have been

```
const { scope } = authorizationRequest
let { authorizationDetails } = authorizationRequest
```

PR #134, opened 2024-07-31
fix: jwk thumprint using crypto.subtle

TimoGlastra:
This PR builds on top of #131 as it is dependant on the changes from dpop.

When testing DPoP in react native it was giving errors for us as crypto.subtle.digest was not defined. So this PR adds the custom hasher implementation to the issuer lib as well, and extends the usage in the RP/OP. Now when you want to use jwk thumprints (for dpop of id-token) you need to provide a hasher.

We could provide a default implementation when in Node.JS / browser and it hasn't been passed, but I didn't do that at the moment.

You can view the changes of this PR here: https://github.com/Sphereon-Opensource/OID4VC/commit/8d869eeea23f1db56b21d1b0211b8abd1f88100f


TimoGlastra:
I think I'm going to abandon this approach as passing the hasher around everywhere is getting very cumbersome. I see in places sha.js is used, so I think I'll just update the usage of crypto.subtle to use sha.js.

What do you think @nklomp

TimoGlastra:
Superseded by #135

PR #135, opened 2024-07-31
fix: jwk thumprint using crypto.subtle

TimoGlastra:
Supersedes #134 (still dependant on #131)

Simpler approach using a default for now. We can later do the registered singleton / service, but as you suggested @nklomp having a default makes it easer to get started.

I also registered the default on the RP/OP so that you now don't have to provide it anymore to use SD-JWT :)


TimoGlastra:
See changes here: https://github.com/Sphereon-Opensource/OID4VC/pull/135/commits/56a291c2a59c2966fdf428d7cf7e2e69389fd38b

PR #142, opened 2024-08-19
fix: client_id_scheme & default scope handling

auer-martin:
A small PR to fix the current client_id_scheme and default scope scope handling.

* Removes the client_id_scheme from the client_metadata (should not be there)
* Removes the cliend_id_scheme from the jwtIssuerX5C (because we also need the client_id_scheme when x5c is not used)
* Add builder method for setting the client_id_scheme
* Remove the openid default scope, it is not required.


nklomp (approved):

nklomp:
LGTM thanks

PR #145, opened 2024-09-05
fix: changes for oid4vc conformance tests

TimoGlastra:
Some fixes that were needed to pass the OID4VC conformance test suite (we are now passing).


sanderPostma (changes requested):
minor refactor to make the code cleaner / more readable

PR #147, opened 2024-09-10
feat: jarm 

auer-martin:
Some left todos.

- Either
  - Allow passing in the encrypted response as string to the lib and decrypt inside the vp library
  - Decrypt outside
- Move MdocVerifiablePresentation to the SSI SDK. (Should not be a class) 
The version is currently not inferred correctly. This is probably due to a type issue with the MDOC presentation definition input.
- No Revocation Verification for Mdoc
- Nonce handling
- Authorization Response Encryption key extraction. (Maybe we should just allow passing in the jwk instead of extracting the encryption jwk from the metadata)
- Pex Integration


TimoGlastra:
> This is going to clash with our mdoc support branch which is soon to be merged.

I think it's best if we update to your mdoc support then as the todo comments also mentions "Move MdocVerifiablePresentation to the SSI SDK. (Should not be a class)". So if you added the needed types to the ssi-sdk we can cross that one off :)

TimoGlastra:
> Either
> - Allow passing in the encrypted response as string to the lib and decrypt inside the vp library
> - Decrypt outside

I think we should allow passing in the encrypted response. Same like we allow passing the signed request.

> - Authorization Response Encryption key extraction. (Maybe we should just allow passing in the jwk instead of extracting the encryption jwk from the metadata)

I think for the Funke we need to extract it from `jwks` in the client_metadata. So maybe we can allow to pass a key, but if it's not provided we  try to extract it from the metadata? I think for most normal cases it'll then work out of the box, as long as you provide a decryption callback.

> Pex Integration

@sanderPostma I saw you also had some MDOC integration work in the PEX library. Is the idea that PEX will support MDOC natively, and that it can also process e.g. a submission including both mdoc and sd-jwt presentations?

packages/siop-oid4vp/lib/op/OP.ts
```diff
@@ -178,6 +195,48 @@ export class OP {
...
+      }
+
+      if (!this._createResponseOptions.encryptJwtCallback) {
+        throw new Error(`Sending an authorization response with response_mode '${responseMode}' requires providing an encryptJwtCallback`)
+      }
+
+      // The client metadata will be parsed in the joseExtractJWKS function
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const jwks = await joseExtractJWKS(clientMetadata as any)
+      const dectyptionJwk = jwks.keys.find((key) => key.use === 'enc')
```
jcmelati, line 211: dec**t**yptionJwk -> dec**r**yptionJwk

auer-martin:
I will start moving the external department here. Are you okay with the general abstractions?

packages/siop-oid4vp/lib/rp/RP.ts
```diff
@@ -133,6 +140,32 @@ export class RP {
     })
   }
 
+  static async processJarmAuthorizationResponse(
+    response: string,
+    opts: {
+      decryptCompact: (input: {
+        jwk: { kid: string }
```
nklomp, line 147: Is there a need for the kid to be required? Kids are application specific and there is no guarantee they are present

nklomp (commented):
See remark about kid in the callback interface. Rest looks okay to proceed

packages/siop-oid4vp/lib/rp/RP.ts
auer-martin: This is from Jarm. It's the 'default' approach.
I can also make the kid optional however, but I think for now we will always use this approach.

(OPTIONAL) The client decrypts the JWT using the key determined by the kid JWT header parameter. The key might be a private key, where the corresponding public key is registered with the expected issuer of the response ("use":"enc" via the client's metadata jwks or jwks_uri) or a key derived from its client secret (see section 4.2).

TimoGlastra:
It seems this test is failing @auer-martin 

> succeed in a full flow with the client using OpenID4VCI version 11 and jwt_vc_json

packages/jarm/lib/jarm-auth-response/v-jarm-auth-response-params.ts
```diff
@@ -0,0 +1,62 @@
+import { checkExp } from '@sphereon/oid4vc-common';
+import * as v from 'valibot';
+
+export const vJarmAuthResponseErrorParams = v.looseObject({
+  error: v.string(),
+  state: v.optional(v.string()),
+
+  error_description: v.pipe(
+    v.optional(v.string()),
+    v.description('Text providing additional information, used to assist the client developer in understanding the error that occurred.'),
```
nklomp, line 10: Are these replaced by actual descriptions, otherwise it needs to be updated

packages/jarm/lib/jarm-auth-response/v-jarm-auth-response-params.ts
```diff
@@ -0,0 +1,62 @@
...
+export const vJarmAuthResponseParams = v.looseObject({
+  state: v.optional(v.string()),
+
+  /**
+   * The issuer URL of the authorization server that created the response
+   */
+  iss: v.string(),
+
+  /**
+   * The client_id of the client the response is intended for
```
nklomp, line 30: Expiration of the JWT

packages/jarm/lib/jarm-auth-response/v-jarm-auth-response-params.ts
```diff
@@ -0,0 +1,62 @@
...
+   */
+  exp: v.number(),
+
+  /**
+   * Expiration of the JWT
```
nklomp, line 35: The audience of the response

packages/jarm/lib/jarm-auth-response/v-jarm-auth-response-params.ts
```diff
@@ -0,0 +1,62 @@
...
+  authResponseParams: JarmAuthResponseParams;
+}) => {
+  const { authRequestParams, authResponseParams } = input;
+  // 2. The client obtains the state parameter from the JWT and checks its binding to the user agent. If the check fails, the client MUST abort processing and refuse the response.
+  if (authRequestParams.state !== authResponseParams.state) {
+    throw new Error(`State missmatch in jarm-auth-response. Expected '${authRequestParams.state}' received '${authRequestParams.state}'.`);
+  }
+
+  // 4. The client obtains the aud element from the JWT and checks whether it matches the client id the client used to identify itself in the corresponding authorization request. If the check fails, the client MUST abort processing and refuse the response.
+  if (authRequestParams.client_id !== authResponseParams.client_id) {
```
nklomp, line 53: This checks client_id against client_id not aud

packages/jarm/lib/jarm-auth-response-send/jarm-auth-response-send.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  switch (responseMode) {
+    case 'direct_post.jwt':
+      return handleDirectPostJwt(responseEndpoint, authResponse);
+    case 'query.jwt':
+      return handleQueryJwt(responseEndpoint, authResponse);
+    case 'fragment.jwt':
+      return handleFragmentJwt(responseEndpoint, authResponse);
+    case 'form_post.jwt':
+      throw new Error('Not implemented');
```
nklomp, line 45: Suggest to include form_post.jwt in the error

packages/siop-oid4vp/lib/__tests__/TestUtils.ts
```diff
@@ -3,6 +3,7 @@
 import crypto, { createHash } from 'crypto'
 
 import { JwtPayload, parseJWT, SigningAlgo, uuidv4 } from '@sphereon/oid4vc-common'
+import { PartialSdJwtDecodedVerifiableCredential } from '@sphereon/pex/dist/main/lib'
```
nklomp, line 6: Import path issue (was already present by the looks of it)

packages/siop-oid4vp/lib/op/OP.ts
```diff
@@ -154,19 +158,47 @@ export class OP {
     }
   }
 
+  public static async extractEncJwksFromClientMetadata(clientMetadata: JwksMetadataParams) {
+    // The client metadata will be parsed in the joseExtractJWKS function
+    const jwks = await extractJwksFromJwksMetadata(clientMetadata)
+    const encryptionJwk = jwks?.keys.find((key) => key.use === 'enc')
```
nklomp, line 164: Probably wise to add a TODO for potential error handling or a warning if multiple JWKs with 'enc' usage are found

nklomp (changes requested):
Some small remarks. The client_id aud validation seems to be incorrect

packages/jarm/lib/jarm-auth-response/v-jarm-auth-response-params.ts
auer-martin: I am not entirely sure what you mean by actual descriptions. 
But the descriptions are not shown to the user if that is what you mean. It's more about attaching them as metadata to the fields.

packages/siop-oid4vp/lib/__tests__/TestUtils.ts
auer-martin: The import doesn't work for me from '@sphereon/pex'.

nklomp: That explains it being there I guess ;). Will fix it in PEX

packages/jarm/lib/jarm-auth-response/v-jarm-auth-response-params.ts
nklomp: Okay, probably should look into valibot. The text suggests an example from their repo; I would expect the description to be not so generic

TimoGlastra:
Is this ready to merge now @nklomp?

nklomp:
Yes, needed to do a round of testing

PR #149, opened 2024-09-25
feat: added schemaValidation to CredentialRequestClient

sksadjad:
added schemaValidation parameter to the CredentialRequestClient(Builder). This parameter is an enum with 3 values: `ALWAYS`, `NEVER`, ` WHEN_PRESENT`. When it's not configured in the builder (with available functions), we default to `WHEN_PRESENT`.


nklomp (commented):
Why are we adding schema validation options to a lib responsible for transport of credentials? The fact we are doing nothing with these options should also be a tell-tale sign. 

Schema validation is credential format specific, not transport specific. So it should not live here

sksadjad:
created the ticket https://sphereon.atlassian.net/browse/SDK-33 for supporting callback function for oid4vp and oid4vci flows

PR #150, opened 2024-09-25
chore: fixes during conformance testing

sanderPostma:


packages/siop-oid4vp/lib/types/Events.ts
```diff
@@ -26,21 +26,27 @@ export enum AuthorizationEvents {
...
   private readonly _timestamp: number
   private readonly _correlationId: string
 
-  public constructor(args: { correlationId: string; subject?: T; error?: Error }) {
+  public constructor(args: { correlationId: string; subject?: T; redirectUri?: string, error?: Error }) {
     //fixme: Create correlationId if not provided. Might need to be deferred to registry though
     this._correlationId = args.correlationId
     this._timestamp = Date.now()
     this._subject = args.subject
+    this._redirectUri = args.redirectUri
```
nklomp, line 39: Huh, we have a generic subject. Why are we creating something very specific (a redirectUri) on every event type. That should not happen

packages/siop-oid4vp/lib/types/VpJwtVerifier.ts
```diff
@@ -96,10 +96,10 @@ export const getRequestObjectJwtVerifier = async (
     if (jwt.payload.redirect_uri && jwt.payload.redirect_uri !== clientId) {
```
nklomp, line 96: There probably should be a else if here, that does the exact same check but then for jwt.payload,response_uri (the new name) instead of redirect_uri

packages/siop-oid4vp/lib/types/VpJwtVerifier.ts
```diff
@@ -96,10 +96,10 @@ export const getRequestObjectJwtVerifier = async (
...
       throw new Error(SIOPErrors.INVALID_CLIENT_ID_MUST_MATCH_REDIRECT_URI)
     }
-    const parts = options.raw.split('.')
+    /*const parts = options.raw.split('.')  this can be signed and execution shouldn't even be here when alg = none
```
nklomp, line 99: Why are we removing this, instead of not checking it when alg === none ?

packages/siop-oid4vp/lib/types/VpJwtVerifier.ts
sanderPostma: Is client_id_scheme still redirect_uri in case  jwt.payload,response_uri is used?
I don't see redirect_uri as valid value for client_id_scheme

packages/siop-oid4vp/lib/types/VpJwtVerifier.ts
sanderPostma: Execution never gets here if alg is none
line 83
```
  if (!clientIdScheme || jwt.header.alg === 'none') {
    return getJwtVerifierWithContext(jwt, { type })
  }
```

packages/siop-oid4vp/lib/types/Events.ts
sanderPostma: It's optional and it's not every event type, just AuthorizationEvent types.
```
await this.emitEvent(AuthorizationEvents.ON_AUTH_RESPONSE_SENT_SUCCESS, { correlationId, subject: response, redirectUri: redirectUri })
```
Alternatively I can create a new type wrapping response & redirectUri. That would be a breaking change, but I do not think we are using that event anywhere else in our codebase.

sanderPostma: I just saw there is a simpler solution without using the event. Will revert

packages/siop-oid4vp/lib/types/VpJwtVerifier.ts
sanderPostma: > There probably should be a else if here, that does the exact same check but then for jwt.payload,response_uri (the new name) instead of redirect_uri

This was redolved BTW

PR #157, opened 2024-10-04
fix: pnpm-lock

auer-martin:
The ci fails currently when publishing the packages.

```
lerna ERR! EUNCOMMIT Working tree has uncommitted changes, please commit or remove the following changes before continuing:
lerna ERR! EUNCOMMIT  M pnpm-lock.yaml
lerna ERR! EUNCOMMIT ?? .npmrc
```

I am unsure, but I think this may have been caused by a different PNPM version I used.


TimoGlastra (approved):
Yeah the pnpm lock version is still on v6, which is quite old. Would be good to upgrade to a newer version

TimoGlastra:
@nklomp bump on this one

nklomp:
We are using pnpm 8 internally, because 9 was not stable at that point. Given the impact on the other repo's and CI/CD this needs to be concerted. So unlikely we will be able to do that today

TimoGlastra:
@auer-martin can you update this to pnpm 8? Or at least the same version as the ci, then it shouldn't cause issues

auer-martin:
I did update to version 8 for compatibility, right? So this should work?

TimoGlastra:
Ah wait, okay this reverts the 9.0 lockfile back to 8. Okay then we should merge this yes.

PR #159, opened 2024-10-10
fix: offer creation improvements

TimoGlastra:
Some improvements / cleanup to the credential offer creation process:
- Take the input  grants as leading, and only set default values if a required prop is not provided. Currently the grants objects are overwritten in some places, which means that a) you can't add custom properties and b) the `authorization_server` in e.g. the pre auth flow was removed when the offer was created
- do not put the whole credential offer uri in the `credential_offer_uri` param, this was leading to weird behaviour where sometimes the value would be the whole uri `openid-credential-offer://credential_offer_uri=<uri>` and sometimes just `<uri>`. So now the object just contains the uri that hosts the credential offer, not the uri that is scanned by the wallet (it's quite confusing that the credential offer uri can contain a `credential_offer_uri` param, so it was somtimes interchanged).
- Correctly encode the `credential_offer_uri` param (Fixes https://github.com/Sphereon-Opensource/OID4VC/issues/102)

### DIFF
--- a/packages/issuer/lib/functions/CredentialOfferUtils.ts
+++ b/packages/issuer/lib/functions/CredentialOfferUtils.ts
@@ -1,21 +1,80 @@
 import { uuidv4 } from '@sphereon/oid4vc-common'
 import {
+  AssertedUniformCredentialOffer,
   CredentialIssuerMetadataOpts,
   CredentialIssuerMetadataOptsV1_0_13,
   CredentialIssuerMetadataV1_0_11,
   CredentialOfferPayloadV1_0_11,
   CredentialOfferPayloadV1_0_13,
   CredentialOfferSession,
-  CredentialOfferV1_0_11,
   CredentialOfferV1_0_13,
   Grant,
+  GrantAuthorizationCode,
   GrantUrnIetf,
   IssuerMetadataV1_0_13,
   PIN_NOT_MATCH_ERROR,
   PRE_AUTH_GRANT_LITERAL,
-  TxCode,
   UniformCredentialOffer,
 } from '@sphereon/oid4vci-common'
+
+export interface CredentialOfferGrantInput {
+  authorization_code?: Partial<GrantAuthorizationCode>
+  [PRE_AUTH_GRANT_LITERAL]?: Partial<GrantUrnIetf>
+}
+
+function createCredentialOfferGrants(inputGrants?: CredentialOfferGrantInput) {
+  // Grants is optional
+  if (!inputGrants || Object.keys(inputGrants).length === 0) return undefined
+
+  const grants: Grant = {}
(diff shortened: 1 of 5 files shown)


TimoGlastra:
Hey @nklomp do you have a chance to look at this PR this week?

packages/issuer/lib/functions/CredentialOfferUtils.ts
```diff
@@ -1,59 +1,96 @@
...
 } from '@sphereon/oid4vci-common'
 
+export interface CredentialOfferGrantInput {
+  authorization_code?: Partial<GrantAuthorizationCode>
+  [PRE_AUTH_GRANT_LITERAL]?: Partial<GrantUrnIetf>
+}
+
+function createCredentialOfferGrants(inputGrants?: CredentialOfferGrantInput) {
+  // Grants is optional
+  if (!inputGrants || Object.keys(inputGrants).length === 0) return undefined
```
nklomp, line 27: We always use accolades to prevent bugs in case someone had a rough night ;)

packages/issuer/lib/functions/CredentialOfferUtils.ts
```diff
@@ -1,59 +1,96 @@
...
+
+  const grants: Grant = {}
+  if (inputGrants?.[PRE_AUTH_GRANT_LITERAL]) {
+    const grant = {
+      ...inputGrants[PRE_AUTH_GRANT_LITERAL],
+      'pre-authorized_code': inputGrants[PRE_AUTH_GRANT_LITERAL]['pre-authorized_code'] ?? uuidv4(),
+    }
+
+    if (grant.tx_code) {
+      if (!grant.tx_code.length) grant.tx_code.length = 4
```
nklomp, line 37: Accolades

packages/issuer/lib/functions/CredentialOfferUtils.ts
```diff
@@ -64,43 +101,15 @@ export function createCredentialOfferObject(
...
-    }
-  } else if (!credential_offer.grants?.authorization_code?.issuer_state) {
-    credential_offer.grants = {
-      authorization_code: {
-        issuer_state: opts?.issuerState ?? uuidv4(),
-      },
-    }
-  }
-  return { credential_offer, credential_offer_uri, scheme, baseUri, grants: credential_offer.grants }
+  if (grants) credential_offer.grants = grants
```
nklomp, line 109: Accolades

nklomp (changes requested):
Looks good. We always use accolades to prevent bugs in case someone had a rough night ;)

nklomp:
@TimoGlastra Small nudge, since you requested for these PRs to be merged

TimoGlastra:
done

nklomp (approved):

PR #170, opened 2024-11-20
Feature/sdk 39 well known oidf (SDK-44 also)

sanderPostma:


sanderPostma:
Also contains a fix for mdoc which i stumbled upon

zoemaas (approved):
LGTM

TimoGlastra:
@sanderPostma curious why did you add this code to bypass PEX for mdocs? I recently added the missing pieces for mdoc in PEX (and we're using mdoc with PEX now). 

```ts
// Handle mdocs, keep them out of pex
  let presentationsArray = (Array.isArray(presentations) ? presentations : [presentations])
  if (presentationsArray.every(p => p.format === 'mso_mdoc')) {
    return
  }  
  presentationsArray = presentationsArray.filter((p) => p.format !== 'mso_mdoc')
```

Is there something we now need to do manually on top of pex to make it work? Won't this make the PD validation fail if it contains mdoc input descriptors?

sanderPostma:
I have been told PEX does not support mdoc a while ago, and because of this our code base has many places where we keep mdoc out of PEX.
In this instance the Funke mdoc PID crashed our Funke RP flow in validatePresentationAgainstDefinition / assertValidVerifiablePresentations.

TimoGlastra:
Can you share the pd and cred?  There's tests in PEX that cover the PID in mdoc format:

https://github.com/Sphereon-Opensource/PEX/blob/develop/test/Mdoc.spec.ts

Which version of PEX are you using?

I think the easiest approach is to use PEX for mdoc as well, as it removes the need for a lot of hacks/workarounds

cc @nklomp

PR #184, opened 2025-08-19
feat/SSISDK-18

BtencateSphereon:


packages/siop-oid4vp/lib/authorization-response/Dcql.ts
```diff
@@ -48,27 +43,67 @@ export class Dcql {
...
+        switch (p.format) {
+          case 'mso_mdoc':
+            return [queryId, {
+                credential_format: p.format,
+                doctype: p.vcs[0].credential.toJson().docType,
+                namespaces: p.vcs[0].decoded,
+                cryptographic_holder_binding: hasCryptographicHolderBinding(p.format, p.vcs[0])
+              } satisfies DcqlMdocCredential
+            ]
+          case 'vc+sd-jwt':
```
nklomp: dc+sd-jwt

packages/siop-oid4vp/lib/authorization-response/Dcql.ts
```diff
@@ -48,27 +43,67 @@ export class Dcql {
...
+            return [queryId, {
```
nklomp: Hmzz, this cannot be right. We created this as a hack for DCQL it seems. We had full support for multiple credentials per presentation using PE. The current code always assumes the first credential even if a presentation has multiple credentials. We should create a FIXME to address this soon

packages/siop-oid4vp/lib/request-object/Payload.ts
```diff
@@ -53,9 +50,7 @@ export const createRequestObjectPayload = async (opts: CreateAuthorizationReques
     state,
     ...registration.payload,
     claims,
-    ...(payload.presentation_definition_uri && { presentation_definition_uri: payload.presentation_definition_uri }),
-    ...(payload.presentation_definition && { presentation_definition: payload.presentation_definition }),
-    ...(payload.dcql_query && { dcql_query: payload.dcql_query }),
+    dcql_query: payload.dcql_query,
```
nklomp: Should be optional, to support SIOPv2

packages/siop-oid4vp/lib/types/SIOP.types.ts
```diff
@@ -714,6 +688,7 @@ export enum SupportedVersion {
   SIOPv2_D11 = 110,
   SIOPv2_D12_OID4VP_D18 = 180,
   SIOPv2_D12_OID4VP_D20 = 200,
+  SIOPv2_V1 = 1000,
```
nklomp: Suggest to rename this one to OID4VP_v1. The current version does not really take compat with SIOPv2 spec into account strictly, and the V1 is really about the OpenID4VP spec, not the draft siopv2 spec

packages/siop-oid4vp/lib/types/VpJwtVerifier.ts
```diff
@@ -78,34 +79,32 @@ export const getRequestObjectJwtVerifier = async (
...
-  if (!clientIdScheme || jwt.header.alg === 'none') {
+  // If a : character is not present in the Client Identifier, the Wallet MUST treat the Client Identifier as referencing a pre-registered client
+  if (!clientIdentifierPrefix || jwt.header.alg === 'none') {
+    // All validations must be done manually
+    // The Verifier metadata is obtained using [RFC7591] or through out-of-band mechanisms.
     return getJwtVerifierWithContext(jwt, { type })
   }
 
-  if (clientIdScheme === 'did') {
+  if (clientIdentifierPrefix === ClientIdentifierPrefix.DECENTRALIZED_IDENTIFIER) {
```
nklomp: or clientId starts with did:. Remember that for know clients to the RP, the prefixes are not mandatory. So a known DID, still should use the DID jwt verifier

nklomp (commented):
Some quick initial comments. Will do a more thorough review later

packages/siop-oid4vp/lib/authorization-response/Dcql.ts
BtencateSphereon: this comes from the ssi-types, do you want me to update there as well? because that will probably hit a lot of things.

or has this been addressed in your ssi branch?

packages/siop-oid4vp/lib/request-object/Payload.ts
BtencateSphereon: is now optional

packages/siop-oid4vp/lib/types/SIOP.types.ts
BtencateSphereon: ranamed

packages/siop-oid4vp/lib/types/VpJwtVerifier.ts
BtencateSphereon: fixed, and made some changes to how we determine the prefix

packages/siop-oid4vp/lib/authorization-response/Dcql.ts
BtencateSphereon: Marking this just as TODO

packages/siop-oid4vp/lib/authorization-response/Dcql.ts
BtencateSphereon: I made a quick fix to include all the credentials for the presentation. Let me know if this is sufficient

packages/siop-oid4vp/lib/authorization-response/Dcql.ts
```diff
@@ -16,65 +26,87 @@ import { extractDcqlPresentationFromDcqlVpToken } from './OpenID4VP'
...
-              credential_format: 'vc+sd-jwt',
-              vct: p.vcs[0].decoded.vct,
-              claims: p.vcs[0].decoded,
-            } satisfies DcqlSdJwtVcCredential,
-          ]
-        } else {
-          throw new Error('DcqlPresentation atm only supports mso_mdoc and vc+sd-jwt')
-        }
-      }),
+        Object.entries(extractDcqlPresentationFromDcqlVpToken(record, opts)).map(([queryId, p]) => {
```
nklomp, line 51: We need to extend this in the future, because a DCQL query can also be about credential sets and thus result in multiple presentations. Current code assumes one presentation. Suggest to create a ticket for that

nklomp (approved):
Approving, because I think it makes more sense to get actual experience with the changed code. Made a comment about single presentation assumption, which we should fix in the near future

packages/siop-oid4vp/lib/authorization-response/Dcql.ts
BtencateSphereon: added FIXME with ticket

packages/siop-oid4vp/lib/authorization-response/Dcql.ts
BtencateSphereon: changed

PR #185, opened 2025-08-21
feat/SSISDK-34

BtencateSphereon:


packages/siop-oid4vp/lib/authorization-response/Dcql.ts
```diff
@@ -71,39 +92,39 @@ export class Dcql {
...
-      vct: vc.decoded.vct,
-      claims: vc.decoded,
+      vct: vc.decodedPayload.vct,
+      claims: vc.decodedPayload,
       cryptographic_holder_binding: hasCryptographicHolderBinding('dc+sd-jwt', vc),
     } satisfies DcqlSdJwtVcCredential
   }
 
-  static toDcqlJwtCredential = (vc: WrappedW3CVerifiableCredential): DcqlW3cVcCredential => {
+  static toDcqlJwtCredential = (vc: JwtDecodedVerifiableCredential): DcqlW3cVcCredential => {
```
nklomp: The previous one accepted any W3C VCDM 1.1 credential. The current one only accepts Jwt Decoded. Is there a specific reason for it?

nklomp (commented):
One question. Is there a specific reason to move away from the wrapped credentials?

packages/siop-oid4vp/lib/authorization-response/Dcql.ts
BtencateSphereon: reverted this, using wrappped again

PR #199, opened 2025-09-18
feature/SSISDK-57_uni-client

sanderPostma:
(Still needs live testing)


packages/siop-oid4vp/lib/types/SIOP.types.ts
```diff
@@ -506,6 +514,11 @@ export enum ResponseMode {
   FRAGMENT_JWT = 'fragment.jwt',
 }
 
+export enum RequestUriMethod {
```
BtencateSphereon: should be a string literal, so i think you need to sync

packages/siop-oid4vp/lib/rp/RP.ts
```diff
@@ -93,6 +96,12 @@ export class RP {
     responseURIType?: ResponseURIType
   }): Promise<AuthorizationRequest> {
     const authorizationRequestOpts = this.newAuthorizationRequestOpts(opts)
+
+    if(opts.queryId && this._dcqlQueryLookupCallback) {
+      const dcqlQuery:DcqlQuery = await this._dcqlQueryLookupCallback(opts.queryId)
```
BtencateSphereon, line 101: should we not do any error handling?

BtencateSphereon (changes requested):

packages/siop-oid4vp/lib/rp/RP.ts
sanderPostma: Why? Any error will cause a Promise.reject

packages/siop-oid4vp/lib/types/SIOP.types.ts
sanderPostma: I am synched up
image

const requestUriMethods = [RequestUriMethod.GET, RequestUriMethod.POST] as const satisfies Array<RequestUriMethod>
export const RequestUriMethodSchema = z.enum(requestUriMethods)

BtencateSphereon: it is a string literal type, so no, you are not synced up i think

export type RequestUriMethod = 'get' | 'post'

sanderPostma: I am synched up with remote feature/DIIPv4. Don't know what to say

BtencateSphereon: ok can you just read what i am typing. because now we get into a discussion which is getting no where

you synced up with code i made and already fixed, that enum is not in the latest DIIPv4 branch, because i know and checked. we have

export type RequestUriMethod = 'get' | 'post'

you are adding code which should not be added......

BtencateSphereon (approved):

PR #208, opened 2025-10-16
feature/SSISDK-73_well-known-draftv1

sanderPostma:


packages/issuer-rest/lib/OID4VCIServer.ts
```diff
@@ -175,7 +175,17 @@ export class OID4VCIServer {
       opts.asClientOpts || this._issuer.asClientOpts ? ({ ...opts.asClientOpts, ...this._issuer.asClientOpts } as ClientMetadata) : undefined
 
     pushedAuthorizationEndpoint(this.router, this.issuer, this.authRequestsData)
-    getMetadataEndpoints(this.router, this.issuer)
+
+    // Create root router for alternative .well-known endpoints if needed
+    const basePath = getBasePath(this.baseUrl)
+    let rootRouter: express.Router | undefined
```
nklomp, line 189: I think the dual hosting should have an actual config option to enable/disable (enabled by default)

nklomp (changes requested):

packages/issuer-rest/lib/OID4VCIServer.ts
sanderPostma: fixed

sanderPostma: Done

PR #216, opened 2025-12-10
feature/SSISDK-94

sanderPostma:


packages/issuer/lib/VcIssuer.ts
```diff
@@ -675,16 +676,70 @@ export class VcIssuer {
...
-      const jwtVerifyResult = jwtVerifyCallback
-        ? await jwtVerifyCallback(credentialRequest.proof)
-        : // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-          await this._jwtVerifyCallback!(credentialRequest.proof)
+      const verifyFn: JWTVerifyCallback | undefined = jwtVerifyCallback ?? this._jwtVerifyCallback
+      if (typeof verifyFn !== 'function') {
+        throw Error(JWT_VERIFY_CONFIG_ERROR)
+      }
+
+      const credReq = credentialRequest as CredentialRequestV1_0_15
```
BtencateSphereon, line 686: do we not have some methods that can check if an object is of a certain type? At some point these casts will fail. We should also have a CredentialRequest type. can that not be used?

packages/issuer/lib/VcIssuer.ts
```diff
@@ -675,16 +676,70 @@ export class VcIssuer {
...
+      }
+
+      // Execute verification
+      let jwtVerifyResult: JwtVerifyResult | undefined
+      const validationErrors: string[] = []
+
+      for (const proof of proofCandidates) {
+        try {
+          jwtVerifyResult = await verifyFn({ jwt: proof.jwt })
+          break
```
BtencateSphereon, line 732: why only the first one? wand what does it mean if we have multiple?

BtencateSphereon (changes requested):

packages/issuer/lib/VcIssuer.ts
sanderPostma: resolved

packages/issuer/lib/VcIssuer.ts
sanderPostma: I am not surewhether to verify all. It's a proof of possession, if we can process & verify one proof we support there is already a a proof of possession.

sanderPostma: Niels Klomp
  [00:07](https://sphereon.slack.com/archives/CHM4EFAQG/p1765494463835289)
One would be okay

So it's fine

BtencateSphereon (approved):

